PR #99, opened 2014-06-11
run go fmt

fbenkstein:
Just run go fmt on the sources.  I guess one should always do that before committing.


carlosmn:
This was done as part of other breaking changes a while ago.

PR #152, opened 2014-12-12
Use a Go package to init OpenSSL

carlosmn:
This let us perform the synchronization at the Go level instead of
hoping nobody else has touched OpenSSL.

---

This would have to wait for spacemonkeygo/openssl#23 to make sure it keeps working on OSX.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ go:
   - 1.3
   - tip
 
+install:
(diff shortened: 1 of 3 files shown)


carlosmn:
Finally giving up on this since OpenSSL is just one of the crypto libraries involved and newer versions are thread-safe by themselves.

PR #164, opened 2015-01-02
Add `Remote#PruneRef` method.

calavera:
And update libgit2 to support prunning.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -69,7 +69,6 @@ type RemoteCallbacks struct {
 	PushUpdateReferenceCallback
 }
 
-
 type Remote struct {
 	ptr       *C.git_remote
 	callbacks RemoteCallbacks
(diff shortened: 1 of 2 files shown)


calavera:
This is blocked by either #152 or https://github.com/libgit2/libgit2/pull/2794 to be able to include `sys/openssl.h`.

remote.go
```diff
@@ -595,6 +592,10 @@ func (o *Remote) UpdateFetchHead() bool {
 	return C.git_remote_update_fetchhead(o.ptr) > 0
 }
 
+func (o *Remote) PruneRefs() bool {
+	return C.git_remote_prune_refs(o.ptr) > 0
+}
```
carlosmn: This is called `C.git_remote_prune()`, we should also call the Go method `Remote.Prune()`.

carlosmn:
You should be able to `#include <git2/sys/opnessl.h>` just under `<git2.h>` and have it work.

carlosmn:
Nevermind, I misread what the openssl include issue was.

calavera:
I'm closing this to start fresh from master, since the openssl issue is fixed there and libgit2 has been also updated.

remote.go
calavera: There are two methods, aren't there? one that retrieves the setting (this call), and another one that performs the prune, the one that you mention. Should I call the setting getter something different?

carlosmn: Oh right, we did leave the getter in; carry on.

PR #165, opened 2015-01-05
Add prune methods to Remote.

calavera:
@carlosmn, I've noticed that `C.git_remote_create` doesn't load the prune configuration. I don't know if that's intentional, if not, I have a patch ready for libgit2.


calavera:
I think this is ready to merge. The changes in libgit2 don't really block this new addition.

PR #166, opened 2015-01-06
Add git note support

benburkert:
Adds support for the `git_note_*` functions.


calavera:
Hi @benburkert :wave: 

Nice PR! :confetti_ball:

benburkert:
@carlosmn I fixed the typo in the comment.

carlosmn:
The receivers for the methods are values rather than objects (e.g. `func (n Note) Id() *Oid {`). AFAICT this will create a copy of the original `Note` object.

This could then create a race condition between this copy for use in these methods and the original object being garbage-collected, which would call the finalizer and free the underlying C object, making the C function call use a dangling pointer.

All this to say that as I understand it, we need to have pointer receivers instead of value receivers.

benburkert:
I switched the methods definitions to pointer receivers. I tend to define methods on values for non mutating functions of small structs (as a way of marking it as "const"), but it looks like the convention in this package is to define methods on pointers.

Although I haven't found anything to back this up definitively, I don't think the finalizer is copied when a struct is placed on the stack. Assuming that's the case, then dangling pointers should not be an issue from inside these methods because the previous stack frame would always hold a reference to the value on the heap and prevent GC from calling the finalizer.

PR #167, opened 2015-01-06
Fix inconsistent function call in Submodule.

calavera:
Keep in mind that this doesn't add the `Update` method that actually performs the update. That's blocked by https://github.com/libgit2/libgit2/pull/2804


calavera:
This should be ready to merge, although there are no tests for submodules :cry:

PR #170, opened 2015-01-22
Add TargetDirectory field to Checkout options.

joseferminj:
TargetDirectory field indicates a alternative checkout path to workdir.


carlosmn:
We need to free the CString after we're done using it, as that's not garbage-collected.

carlosmn:
Merged manually, thanks.

PR #173, opened 2015-02-11
Add bindings for git_graph_* methods

AaronO:
This adds support for `git_graph_ahead_behind` and `git_graph_descendant_of` respectively wrapped as `Repository.GraphAheadBehind` and `Repository.GraphDescendantOf`.

### DIFF
--- a/graph.go
+++ b/graph.go
@@ -1,0 +1,36 @@
+package git
+
+/*
+#include <git2.h>
+*/
+import "C"
+import (
+	"runtime"
+)
+
+func (repo *Repository) GraphDescendantOf(commit, ancestor *Oid) (bool, error) {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ret := C.git_graph_descendant_of(repo.ptr, commit.toC(), ancestor.toC())
+	if ret < 0 {
+		return false, MakeGitError(ret)
+	}
+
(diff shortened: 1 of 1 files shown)


carlosmn:
I think we could do without the "Graph" prefix. `Repository.AheadBehind()` etc does not loose any information.

AaronO:
@carlosmn I considered that, but I ended up thinking that it was probably best to stick to `libgit2`'s naming convention to avoid any confusion.

For example: https://github.com/libgit2/libgit2/issues/1440, there was a discussion about implementing `git_ahead_behind` that returns a list of commits instead of a number, etc ...

If the prefix is unnecessary/redundant, shouldn't the decision to strip it be made at the `libgit2` level and `git2go` should reflect whatever happens there ?

Either way, I can strip the prefix if you think that's best.

AaronO:
@carlosmn Voila, I removed the `Graph` prefix.

carlosmn:
ahead-behind (or left-right in git parlance) isn't a first-class object. It's a question you ask of a pair of commits, so it doesn't get its own namespace. It doesn't fit in the `git_repository` namespace because it's not *about* the repository, so it was shoved to "graph".

But in OO syntax, the receiver is the thing you take as source or destination of data, so in the other bindings, it's gone directly there.

PR #174, opened 2015-02-14
use git_signature_free (fix test crash on Windows)

mdonoughe:
libgit2 0.22.1 compiled x64 RelWithDebug using Visual Studio 2013 combined with the v22 branch crashes when running the unit tests. I've traced this to the use of `C.free(unsafe.Pointer(csig))` in some places instead of `C.git_signature_free(csig)` as in other places. Switching to `git_signature_free` seems to fix this. I don't know why in some places `free` was used over `git_signature_free`.

### DIFF
--- a/reference.go
+++ b/reference.go
@@ -37,7 +37,7 @@ func (v *Reference) SetSymbolicTarget(target string, sig *Signature, msg string)
 	defer runtime.UnlockOSThread()
 
 	csig := sig.toC()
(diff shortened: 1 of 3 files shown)


mdonoughe:
I forgot to say this is go 1.4.1 and gcc is tdm64 4.9.2-3 and I'm running on Windows 8.1.

It looks like the CI configuration for this branch is not right.

PR #176, opened 2015-02-18
Add BranchIterator#ForEach.

calavera:
This abstracts the branch iteration from the user allowing them to define a function to process branch and type.
See test case for an easy example.

### DIFF
--- a/branch.go
+++ b/branch.go
@@ -65,7 +65,7 @@ func (i *BranchIterator) Free() {
 func (i *BranchIterator) ForEach(f BranchIteratorFunc) error {
 	b, t, err := i.Next()
 
-	for b != nil && err == nil {
+	for err == nil {
 		err = f(b, t)
 		if err == nil {
 			b, t, err = i.Next()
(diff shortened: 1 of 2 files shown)


calavera:
@carlosmn any comments about this? I think it's more user friendly than having to call `Next` yourself.

branch.go
```diff
@@ -65,8 +62,20 @@ func (i *BranchIterator) Free() {
 	C.git_branch_iterator_free(i.ptr)
 }
 
-func (repo *Repository) NewBranchIterator(flags BranchType) (*BranchIterator, error) {
+func (i *BranchIterator) ForEach(f BranchIteratorFunc) error {
+	b, t, err := i.Next()
+
+	for b != nil && err == nil {
```
carlosmn: `b` shouldn't ever be `nil` in a situation that you can't decide by checking if `err != nil`, I don't see what that check adds here.

calavera:
Thanks for the comment @carlosmn. I've removed that unnecessary check.

carlosmn:
I don't quite understand the dislike of for loops that seems so prevalent among Go programmers, but sure. I've squashed and pushed manually. Thanks.

PR #180, opened 2015-03-14
Add possibiliy of checkout on specific path

arcamael:


checkout.go
```diff
@@ -65,6 +66,15 @@ func populateCheckoutOpts(ptr *C.git_checkout_options, opts *CheckoutOpts) *C.gi
...
 	}
+	if len(opts.Paths) > 0 {
+		var pathArray []*C.char = make([]*C.char, len(opts.Paths))
+		for i, s := range opts.Paths {
+			pathArray[i] = C.CString(s)
+		}
+		ptr.paths.strings = (**C.char) (unsafe.Pointer(&pathArray[0]))
+		ptr.paths.count = C.size_t(len(opts.Paths))
+	}
+
```
carlosmn: We have a function so we don't have to write this code every time, use `makeCstringsFromString()`.

checkout.go
```diff
@@ -73,6 +83,13 @@ func freeCheckoutOpts(ptr *C.git_checkout_options) {
 		return
 	}
 	C.free(unsafe.Pointer(ptr.target_directory))
+	if int(ptr.paths.count) > 0 {
+		var strings **C.char = ptr.paths.strings
+		slice := (*[1 << 30]*C.char)(unsafe.Pointer(strings))[0:int(ptr.paths.count)]
+		for _, s := range slice {
+			C.free(unsafe.Pointer(s))
+		}
+	}
```
carlosmn: We have `freeStrarray()` to free go-allocated string arrays.

PR #201, opened 2015-05-19
is our test writer drunk?

taylorchu:
very drunk, it seems.

### DIFF
--- a/index_test.go
+++ b/index_test.go
@@ -128,8 +128,6 @@ func checkFatal(t *testing.T, err error) {
 	// The failure happens at wherever we were called, not here
 	_, file, line, ok := runtime.Caller(1)
 	if !ok {
-		t.Fatal()
+		t.Fatalf("Fail at %v:%v; %v", file, line, err)
 	}
-
-	t.Fatalf("Fail at %v:%v; %v", file, line, err)
 }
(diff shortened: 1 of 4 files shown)


reference_test.go
```diff
@@ -199,8 +199,6 @@ func checkRefType(t *testing.T, ref *Reference, kind ReferenceType) {
 	// The failure happens at wherever we were called, not here
 	_, file, line, ok := runtime.Caller(1)
 	if !ok {
-		t.Fatal()
+		t.Fatalf("Wrong ref type at %v:%v; have %v, expected %v", file, line, ref.Type(), kind)
 	}
-
-	t.Fatalf("Wrong ref type at %v:%v; have %v, expected %v", file, line, ref.Type(), kind)
```
carlosmn: This is correct. The `!ok` block is for the case where we are unable to access the stack. This change would make it so `checkRefType` would never assert the type of ref.

index_test.go
```diff
@@ -128,8 +128,6 @@ func checkFatal(t *testing.T, err error) {
 	// The failure happens at wherever we were called, not here
 	_, file, line, ok := runtime.Caller(1)
 	if !ok {
-		t.Fatal()
+		t.Fatalf("Fail at %v:%v; %v", file, line, err)
 	}
-
-	t.Fatalf("Fail at %v:%v; %v", file, line, err)
```
carlosmn: This is correct, the `!ok` block refers to the `runtime.Caller(1)` call, it says nothing about the test.

PR #204, opened 2015-05-19
fix lots of golint warnings

taylorchu:


carlosmn:
This PR changes the public interface, so we can't merge it into master.

But from what I see, the changes are changing 'Id' to 'ID' and 'Url' to 'URL', which seem to be completely cosmetic on internal variables, so it introduces *a lot* of code churn, and I don't really see there's benefit to the changes.

PR #212, opened 2015-06-10
Make the network code use handles

carlosmn:
This wasn't ported together with the rest, but it does exhibit the same
issues, so let's port it over now.


michaeledgar:
This patch is good, and resolves my issues. Thanks!

Sorry it took so long to respond to this - I had a separate bug in my code which presented nearly identically during remote operations. (I was trying to reuse *git.Cred objects, but the underlying *C.git_cred is freed by libgit2 after the first use)

PR #222, opened 2015-07-17
Add more support for tags

clns:
Add `TagsCollection`.

Add support for `git_tag_list`, `git_tag_list_match`, `git_tag_foreach`, `git_tag_create_lightweight`.

### DIFF
--- a/tag.go
+++ b/tag.go
@@ -2,8 +2,14 @@ package git
 
 /*
 #include <git2.h>
+
+extern int _go_git_tag_foreach(git_repository *repo, void *payload);
 */
 import "C"
+import (
+	"runtime"
+	"unsafe"
+)
 
 // Tag
 type Tag struct {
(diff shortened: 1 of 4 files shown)


carlosmn:
This moves the code back to having a lot of methods on `Repository`, which is a style which current git2go has abandoned. These should be methods on a `TagsCollection` object.

tag.go
```diff
@@ -42,3 +48,129 @@ func (t Tag) TargetId() *Oid {
...
+// The name of the tag is validated for consistency (see git_tag_create() for the rules
+// https://libgit2.github.com/libgit2/#HEAD/group/tag/git_tag_create) and should
+// not conflict with an already existing tag name.
+//
+// If force is true and a reference already exists with the given name, it'll be replaced.
+//
+// The created tag is a simple reference and can be queried using
+// repo.References.Lookup("refs/tags/<name>"). The name of the tag (eg "v1.0.0")
+// is queried with ref.Shorthand().
+func (repo *Repository) CreateTagLightweight(name string, commit *Commit, force bool) (*Oid, error) {
```
carlosmn: This needs to explain what the return Oid is, there is no way to know what it is from the signature.

clns: I added more explanation in the next commit.

clns:
I noticed the other collections but since there wasn't any for tags I didn't know if I should add one. I can go ahead and add a `TagsCollection`, let me know.

carlosmn:
Yes, do add `TagsCollection`.

clns:
Added `TagsCollection` and moved the new methods on it.

This PR doesn't break any existing apis, so moving the existing tags-related methods from `Repository` into `TagsCollection` is outside of the scope of this PR and is a todo.

PR #226, opened 2015-07-22
Update to libgit v0.23.0 [WIP/RFC]

pks-t:
This pull request updates git2go to libgit v0.23.0. While it compiles, tests have not been adjusted to API changes yet. Thus this should be considered as work in progress.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ language: go
 
 install:
   - cd "${HOME}"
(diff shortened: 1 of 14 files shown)


carlosmn:
The `next` branch is updated to v0.23, that's what should get merged to master once the issues with the handle tracking are resolved.

pks-t:
Damn, didn't check if there was a branch doing this already. I'll take a try at the handle tracking, then.

PR #227, opened 2015-07-23
Add git-describe support

clns:
Includes `git_describe_commit` and `git_describe_workdir`, with all existing describe and format options.

### DIFF
--- a/describe.go
+++ b/describe.go
@@ -1,0 +1,222 @@
+package git
+
+/*
+#include <git2.h>
+*/
+import "C"
+import (
+	"runtime"
+	"unsafe"
+)
+
+// DescribeOptions represents the describe operation configuration.
+//
+// You can use DefaultDescribeOptions() to get default options.
+type DescribeOptions struct {
+	// How many tags as candidates to consider to describe the input commit-ish.
+	// Increasing it above 10 will take slightly longer but may produce a more
+	// accurate result. 0 will cause only exact matches to be output.
+	MaxCandidatesTags uint // default: 10
+
+	// By default describe only shows annotated tags. Change this in order
+	// to show all refs from refs/tags or refs/.
+	Strategy DescribeOptionsStrategy // default: DescribeDefault
+
+	// Only consider tags matching the given glob(7) pattern, excluding
+	// the "refs/tags/" prefix. Can be used to avoid leaking private
+	// tags from the repo.
+	Pattern string
+
+	// When calculating the distance from the matching tag or
+	// reference, only walk down the first-parent ancestry.
+	OnlyFollowFirstParent bool
+
+	// If no matching tag or reference is found, the describe
+	// operation would normally fail. If this option is set, it
+	// will instead fall back to showing the full id of the commit.
+	ShowCommitOidAsFallback bool
+}
+
+// DefaultDescribeOptions returns default options for the describe operation.
+func DefaultDescribeOptions() (DescribeOptions, error) {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	opts := C.git_describe_options{}
+	ecode := C.git_describe_init_options(&opts, C.GIT_DESCRIBE_OPTIONS_VERSION)
+	if ecode < 0 {
+		return DescribeOptions{}, MakeGitError(ecode)
+	}
+
+	return DescribeOptions{
+		MaxCandidatesTags: uint(opts.max_candidates_tags),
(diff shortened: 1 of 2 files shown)


clns:
Any reason why this can't be merged?

carlosmn:
Not a priori, but it needs to get reviewed, which takes time.

describe.go
```diff
@@ -0,0 +1,204 @@
...
+	opts := C.git_describe_options{}
+	ecode := C.git_describe_init_options(&opts, C.GIT_DESCRIBE_OPTIONS_VERSION)
+	if ecode < 0 {
+		return DescribeOptions{}, MakeGitError(ecode)
+	}
+
+	return DescribeOptions{
+		MaxCandidatesTags:       uint(opts.max_candidates_tags),
+		Strategy:                DescribeOptionsStrategy(opts.describe_strategy),
+		Pattern:                 C.GoString(opts.pattern),
```
carlosmn: The pattern is user-defined. There is no pattern which the initaliser would set.

describe.go
```diff
@@ -0,0 +1,204 @@
...
+		OnlyFollowFirstParent:   opts.only_follow_first_parent == 1,
+		ShowCommitOidAsFallback: opts.show_commit_oid_as_fallback == 1,
+	}, nil
```
carlosmn: And the same with these ones. This isn't going to be true.

describe.go
```diff
@@ -0,0 +1,204 @@
...
+	opts := C.git_describe_format_options{}
+	ecode := C.git_describe_init_format_options(&opts, C.GIT_DESCRIBE_FORMAT_OPTIONS_VERSION)
+	if ecode < 0 {
+		return DescribeFormatOptions{}, MakeGitError(ecode)
+	}
+
+	return DescribeFormatOptions{
+		AbbreviatedSize:     uint(opts.abbreviated_size),
+		AlwaysUseLongFormat: opts.always_use_long_format == 1,
+		DirtySuffix:         C.GoString(opts.dirty_suffix),
```
carlosmn: Here as well, the default is false and `NULL`, or in Go an empty string, these answers won't change.

describe.go
```diff
@@ -0,0 +1,204 @@
...
+
+// Describe strategy options.
+const (
+	DescribeDefault DescribeOptionsStrategy = C.GIT_DESCRIBE_DEFAULT
+	DescribeTags    DescribeOptionsStrategy = C.GIT_DESCRIBE_TAGS
+	DescribeAll     DescribeOptionsStrategy = C.GIT_DESCRIBE_ALL
+)
+
+// Describe performs the describe operation on the commit.
+func (c Commit) Describe(describeOpts *DescribeOptions, formatOpts *DescribeFormatOptions) (string, error) {
```
carlosmn: This should be a pointer receiver. Copying the commit would negate any garbage-collection guarantees we're trying to keep.

describe.go
```diff
@@ -0,0 +1,204 @@
...
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ecode := C.git_describe_commit(&resultPtr, c.gitObject.ptr, cDescribeOpts)
+	if ecode < 0 {
+		return "", MakeGitError(ecode)
+	}
+	defer C.git_describe_result_free(resultPtr)
+
+	return describeFormat(resultPtr, formatOpts)
```
carlosmn: Why force the user to take the string? The operation and its textual output from git are two distinct things.

describe.go
```diff
@@ -0,0 +1,204 @@
...
+	ecode := C.git_describe_workdir(&resultPtr, repo.ptr, cDescribeOpts)
+	if ecode < 0 {
+		return "", MakeGitError(ecode)
+	}
+	defer C.git_describe_result_free(resultPtr)
+
+	return describeFormat(resultPtr, formatOpts)
+}
+
+func describeFormat(resultPtr *C.git_describe_result, formatOpts *DescribeFormatOptions) (string, error) {
```
carlosmn: This function should be exported and either have a `DescribeResult` receiver or take one as a parameter.

describe.go
clns: What do you propose it's best here?

carlosmn: Stick to exposing what libgit2 does. One describe operation which returns a struct and one operation which formats that information like git does.

clns:
Pushed the updated version, check it out when you get a chance, thanks.

PR #229, opened 2015-07-24
Fix formatting

joseferminj:
Fix formatting in remote.go file

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -72,12 +72,12 @@ type RemoteCallbacks struct {
 type FetchPrune uint
 
 const (
-	 // Use the setting from the configuration
+	// Use the setting from the configuration
 	FetchPruneUnspecified FetchPrune = C.GIT_FETCH_PRUNE_UNSPECIFIED
 	// Force pruning on
-	FetchPruneOn       FetchPrune = C.GIT_FETCH_PRUNE
+	FetchPruneOn FetchPrune = C.GIT_FETCH_PRUNE
 	// Force pruning off
-	FetchNoPrune       FetchPrune = C.GIT_FETCH_NO_PRUNE
+	FetchNoPrune FetchPrune = C.GIT_FETCH_NO_PRUNE
 )
 
 type DownloadTags uint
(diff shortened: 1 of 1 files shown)


carlosmn:
This is all whitespace changes. This adds a lot of code churn for very little gain.

joseferminj:
This is go standard format. All this whitespace changes are introduced by 'go fmt' tool. 
This is the output of running 'go fmt remote.go'.

I have pending pull requests that modified remote.go, but before sending them have to format this file according Go standard format, otherwise my changes will get mixed with the formatting and they are difficult to read.

carlosmn:
This is precisely the reason to avoid wholesale formatting changes. Any changes people have will no longer apply without giving us any advantage.

joseferminj:
The problem is that  every time I run 'go fmt' locally, it will modify my
local files to adjust the format, in particular 'remote.go'. The effect is
that formatting change is merged with the actual change I am working on. To
avoid this problem, this pull request make remote.go to follow the Go
standard format.

All files in the project should be formatted following the Go standard
before pushing them to the repository.

carlosmn:
The only thing I can say to that is not to run `go fmt` on a file which is not already in the format the tool produces, or exclude those changes when preparing the commit. If a piece of code isn't being worked on, doing formatting changes causes precisely this problem you're encountering.

clns:
Not running the tool on certain files might prove difficult since a lot of the go editors do this automatically on save (vim-go, atom, etc.). There is already a PR to run `go fmt` on all the files ( #99 ) but it's probably outdated.

I would vote for considering this in the near future, to ease development.

mvdan:
The more time that passes without `go fmt ./...` being run, the bigger this problem will be. You just can't have a project with its own format.

nathanleclaire:
@carlosmn @vmg 

Since the upstream Go project recommends running `gofmt` on Go files and would likely prefer all Go projects to run `gofmt` (https://blog.golang.org/go-fmt-your-code#TOC_2.), why not run `gofmt` on the codebase?  It's dictated by the upstream directly moreso than personal preference.

Since many Go programmers have a editor hook which formats their Go code automatically, are you willing to accept that refusing to `gofmt` the code in this repository will discourage contribution activity as very few will be willing to disable and workaround their default workflow just to contribute to this project?  On fast moving Go projects which require their contributors to `gofmt` code (Docker, Docker Machine, Hashicorp tools) I have very rarely seen developers complain about having to reviewed `gofmt`-ed code, in fact the opposite is usually the case - it makes folks very happy.

How about a compromise where a `gofmt` change is added to the Travis CI check and all pull requests are required to run `gofmt`?  This would ensure that this consistency is enforced always, and reduce the concern of noisy code diffs.  I would be happy to make a PR for such a change and `gofmt` the existing files in the repository as well.  Are you open to that?

carlosmn:
This was done as part of other breaking changes a while ago.

PR #245, opened 2015-08-23
Restore support for RemoteCallbacks in Remote.Push() [next]

AaronO:
The `next` branch dropped support for `RemoteCallbacks` in `Remote.Push()` while still being available for `Remote.Fetch()`, this PR adds them back.

Without them it's impossible to authenticate during a push, etc ...


carlosmn:
Thanks, I picked up the changes removing the whitespace diff.

PR #253, opened 2015-09-16
Expose baseline field in CheckoutOptions

joseferminj:


carlosmn:
Can you squash together the commits? Having one which doesn't build can disrupt bug-hunting.

joseferminj:
done.

PR #257, opened 2015-09-21
[next] Add stash support

clns:


clns:
Can we have an update on this?

clns:
Is there anything that can be done to have this PR reviewed and merged?

clns:
Still nothing?

stash.go
```diff
@@ -0,0 +1,334 @@
...
+	optsC := C.git_stash_apply_options{}
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ecode := C.git_stash_apply_init_options(&optsC, C.GIT_STASH_APPLY_OPTIONS_VERSION)
+	if ecode < 0 {
+		return StashApplyOptions{}, MakeGitError(ecode)
+	}
+	defer freeStashApplyOptions(&optsC)
```
carlosmn: It looks like you're using `optsC` to copy the defaults into the Go struct. In that case, you haven't allocated anything, so there's nothing to free.

stash.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+	return
+}
+
+func freeStashApplyOptions(optsC *C.git_stash_apply_options) {
+	if optsC != nil {
+		freeCheckoutOpts(&optsC.checkout_options)
+		if optsC.progress_payload != nil {
+			pointerHandles.Untrack(optsC.progress_payload)
+		}
```
carlosmn: The untracking would be better suited to a defer or similar around the function that actually uses it. The pattern we have is to use these free functions to free resources allocated with the C allocator, which this is not.

stash.go
```diff
@@ -0,0 +1,334 @@
...
+
+	if opts != nil {
+		progressData = &stashApplyProgressData{
+			Callback: opts.ProgressCallback,
+		}
+
+		optsC = &C.git_stash_apply_options{
+			version:          C.GIT_STASH_APPLY_OPTIONS_VERSION,
+			flags:            C.git_stash_apply_flags(opts.Flags),
+			checkout_options: *opts.CheckoutOptions.toC(),
```
carlosmn: This should be `populateCheckoutOpts`. With this you're doing and undoing the work of getting a pointer to some struct we're going to throw away.

stash.go
clns: I added `untrackStashApplyOptionsCallback()` and moved the untracking there, since it cannot be added to `toC()` where the tracking happens.

For the reference, a lot of options that have callbacks do the same thing, e.g. [clone.go](https://github.com/libgit2/git2go/blob/1c855246ca0cb3faf7d8a644b22355e0ef6856a1/clone.go#L105), [diff.go](https://github.com/libgit2/git2go/blob/1c855246ca0cb3faf7d8a644b22355e0ef6856a1/diff.go#L566).

stash.go
clns: Yeah, I must've copied this from [cherrypick.go](https://github.com/libgit2/git2go/blob/1c855246ca0cb3faf7d8a644b22355e0ef6856a1/cherrypick.go#L47), so it should be fixed there as well.

clns:
Thanks for the review. I made the changes, pushed them and commented on some.

PR #258, opened 2015-09-26
Fix bug in Config LookupString

TheDahv:
A solution to `get_string called on a live config object` errors when attempting a string lookup on a config object.

### DIFF
--- a/config.go
+++ b/config.go
@@ -115,18 +115,20 @@ func (c *Config) LookupInt64(name string) (int64, error) {
 }
 
 func (c *Config) LookupString(name string) (string, error) {
-	var ptr *C.char
 	cname := C.CString(name)
 	defer C.free(unsafe.Pointer(cname))
+
+	valBuf := C.git_buf{}
 
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
-	if ret := C.git_config_get_string(&ptr, c.ptr, cname); ret < 0 {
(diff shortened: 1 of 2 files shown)


TheDahv:
@carlosmn, @vmg - Not sure what the best way to contact you for discussion or questions. Am I just using the Config methods wrong? I kept running into errors when doing something along these lines:

```go
path, _ := ConfigFindGlobal()
conf, _ = OpenOndisk(nil, path)
val, err = c.LookupString("foo.bar")
// err should complain about performing a lookup on a live config object
```

If that's supposed to work, then adding this snapshot method on config objects fixes the problem.

carlosmn:
The snapshotting solves a different problem. We simply should not be using `git_config_get_string()` at all, but `git_config_get_string_buf()`. The former is something you can do if you know the lifetimes of your objects, but that's not something you can really know in Go.

TheDahv:
@carlosmn thanks for the tip and explanation. I updated my approach to use `git_config_get_string_buf()`. I also added some tests for the other Lookup methods. Hopefully they might serve as documentation of example usage for future users of this library.

TheDahv:
@carlosmn any further feedback or shall I just wait until more time frees up for you?

carlosmn:
Just to clean up the history here. This PR still contains commits to add and remove snapshotting capabilities. If all you're aiming for is to get string lookups to work, that's the only one it should have.

TheDahv:
Sorry, got caught up for a few days. Commits cleaned up and squashed.

PR #262, opened 2015-10-09
Add support for ignore

clearr:
This adds support for `git_ignore_*` functions.

### DIFF
--- a/ignore.go
+++ b/ignore.go
@@ -1,0 +1,51 @@
+package git
+
+/*
+#include <git2.h>
+*/
+import "C"
+import (
+	"runtime"
+	"unsafe"
+)
+
+func (v *Repository) AddIgnoreRule(rules string) error {
+	crules := C.CString(rules)
+	defer C.free(unsafe.Pointer(crules))
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ret := C.git_ignore_add_rule(v.ptr, crules)
+	if ret < 0 {
+		return MakeGitError(ret)
+	}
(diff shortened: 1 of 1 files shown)


ignore.go
```diff
@@ -0,0 +1,44 @@
...
+
+/*
+#include <git2.h>
+*/
+import "C"
+import (
+	"runtime"
+)
+
+func (v *Repository) AddInternalIgnoreRules(rules string) error {
```
carlosmn: This name doesn't match the libgit2 name. It should also not be plural, as it only accepts a single rule to add.

clearr: [This document](https://libgit2.github.com/libgit2/#HEAD/group/ignore/git_ignore_add_rule) says that the parameter `rules` can have multiple rules in it.
Anyway, I will change it to `AddIgnoreRule(rules string)` to match libgit2.

ignore.go
```diff
@@ -0,0 +1,44 @@
...
+import "C"
+import (
+	"runtime"
+)
+
+func (v *Repository) AddIgnoreRule(rules string) error {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ret := C.git_ignore_add_rule(v.ptr, C.CString(rules))
```
carlosmn: This (and the one below) leaks the string. You need to keep it in a variable and free it after it's done via `C.free()`, which is usually done in a `defer`. Check out the other places in the library where we do this for examples.

clearr:
I fixed leaks and squashed commits.

PR #268, opened 2015-10-29
Fix memory leaks in NewIndex() and OpenIndex()

### LINKED ISSUE #267: Index memory leaks?
`Index` has [`Free()`](https://github.com/libgit2/git2go/blob/3b5633de219eec2908723a8e557758cc055d84d1/index.go#L317-L320) method and [`newIndexFromC()`](https://github.com/libgit2/git2go/blob/3b5633de219eec2908723a8e557758cc055d84d1/index.go#L82-L86) makes sure the method is called.
`newIndexFromC()` is used in [`Repository.Index()`](https://github.com/libgit2/git2go/blob/3b5633de219eec2908723a8e557758cc055d84d1/repository.go#L145), but not in [`NewIndex()`](https://github.com/libgit2/git2go/blob/3b5633de219eec2908723a8e557758cc055d84d1/index.go#L90-L101) and [`OpenIndex()`](https://github.com/libgit2/git2go/blob/3b5633de219eec2908723a8e557758cc055d84d1/index.go#L105-L119).

I think this causes memory leaks when we use these functions without calling `Index.Free()` explicitly.

clearr:
Fixes #267.

### DIFF
--- a/index.go
+++ b/index.go
@@ -97,7 +97,7 @@ func NewIndex() (*Index, error) {
 		return nil, MakeGitError(err)
 	}
 
-	return &Index{ptr: ptr}, nil
+	return newIndexFromC(ptr), nil
 }
 
 // OpenIndex creates a new index at the given path. If the file does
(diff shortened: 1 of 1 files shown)


carlosmn:
Thanks!

PR #270, opened 2015-11-12
Add Commit.Time()

TrungACZNE:


carlosmn:
This will return a different result from `git show`, `git log` and friends unless you also take the timezone offset into consideration.

harobed:
Up

PR #280, opened 2016-01-04
Don't drop CreateBlobFromChunks hintPath argument

ebfe:


carlosmn:
Man, that `if` block was quite silly. Thanks.

PR #282, opened 2016-01-08
handles, merge, odb: changes for Go 1.6 pointer passing rules

ianlancetaylor:
See http://tip.golang.org/cmd/cgo/#hdr-Passing_pointers .

### DIFF
--- a/handles.go
+++ b/handles.go
@@ -10,15 +10,14 @@ type HandleList struct {
 	sync.RWMutex
 	// stores the Go pointers
 	handles []interface{}
-	// Indicates which indices are in use, and keeps a pointer to slot int variable (the handle)
-	// in the Go world, so that the Go garbage collector does not free it.
-	set map[int]*int
+	// Indicates which indices are in use.
+	set map[int]bool
 }
 
 func NewHandleList() *HandleList {
 	return &HandleList{
 		handles: make([]interface{}, 5),
-		set:     make(map[int]*int),
+		set:     make(map[int]bool),
 	}
 }
 
@@ -26,8 +25,7 @@ func NewHandleList() *HandleList {
 // list. You must only run this function while holding a write lock.
 func (v *HandleList) findUnusedSlot() int {
 	for i := 1; i < len(v.handles); i++ {
-		_, isUsed := v.set[i]
-		if !isUsed {
+		if !v.set[i] {
 			return i
 		}
 	}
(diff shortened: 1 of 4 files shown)


merge.go
```diff
@@ -321,17 +321,21 @@ type MergeFileInput struct {
...
 
 // populate a C struct with merge file input, make sure to use freeMergeFileInput to clean up allocs
-func populateCMergeFileInput(c *C.git_merge_file_input, input MergeFileInput) {
+func populateCMergeFileInput(c *C.git_merge_file_input, input MergeFileInput) *C.char {
 	c.path = C.CString(input.Path)
+	var toFree *C.char
 	if input.Contents != nil {
-		c.ptr = (*C.char)(unsafe.Pointer(&input.Contents[0]))
+		toFree = C.CString(string(input.Contents))
+		c.ptr = toFree
```
carlosmn: This makes us allocate an extra buffer. We also don't actually know that what we have is a string. It's some buffer, so making it go through the string type seems wrong.

If there's no way to work around this in Go, I'd rather we write some C so we're passing `&input.Contents[0]` to a C function so the rules make it OK for us to re-use this buffer.

handles.go
```diff
@@ -52,12 +52,12 @@ func (v *HandleList) Track(pointer interface{}) unsafe.Pointer {
 
 	v.Unlock()
 
-	return unsafe.Pointer(&slot)
+	return unsafe.Pointer(uintptr(slot))
```
carlosmn: This changes what we're doing, doesn't it? We were taking its address and making it a `unsafe.Pointer`. In the new code we're taking the slot value and casting it a few times. Should this still work?

carlosmn:
There's a bunch of unrelated whitespace change that makes this harder to read, could you drop that?

handles.go
carlosmn: Ok, I see why we have to do this. But this means that the `set` map is now not really keeping the pointers alive. It should then move to `map[int]bool` since we only need it to be set.

ianlancetaylor: Done.

ianlancetaylor:
Pull request updated.

I dropped the whitespace changes, but they were entirely from gofmt.  I would suggest running gofmt on the Go files.

carlosmn:
The `next` branch did get a `go fmt` treatment, but I've kept `master` as-is to reduce the amount of in-flight work that would conflict due to whitespace change.

Thanks!

PR #285, opened 2016-01-27
Compile fixes for libgit2 0f9d15493d5d8ad4353dd7beed52c9567334f6e5

hanwen:

PR #289, opened 2016-02-03
Treebuilder signature

hanwen:
the updated types on the constants cause TreeBuilder methods to fail.


carlosmn:
This are the same changes from #285 so I'm closing this. We track the release with master so we won't change function signatures here.

PR #298, opened 2016-03-06
Add git_diff_tree_to_index

hirochachacha:
Hi, I'm not sure about `libgit2`, but I need `git diff --cached` -ish features.

According to https://libgit2.github.com/libgit2/#HEAD/group/diff/git_diff_tree_to_index:

It requires `add_git_diff_tree_to_index`. So I've added it.

Thanks.


carlosmn:
Please remove the `go fmt` commit. It's stopping this from being applyable. Looks good otherwise.

PR #301, opened 2016-03-30
Add some constants and repository methods.

hansrodtang:
Added some tiny stuff I needed. I did this on the next branch, but it looks like some of this is available on previous versions on libgit2 (if i'm not mistaken). Let me know if anything is wrong and I'll fix it quickly!


carlosmn:
Thanks. These can go in either place, so I'll merge it against 'next'. It can be backported if anybody does need it and they're on 'master'.

PR #302, opened 2016-03-31
Fix Go 1.6 incompatiblity with blob creation

itamarst:
In certain situations the double-pointer rules in cgo 1.6 break `CreateBlobFromBuffer`.

This PR:
1. Adds 1.6 to automatically tested list of Go versions for Travis-CI.
2. Adds a couple of tests for `CreateBlobFromBuffer`, one of which triggers this edge case.
3. Provides a (non-ideal) fix.

I have only tested this against the `v23` branch which is what I am using locally. I would appreciate backport to `v23` once this or some other fix is merged to `master`.


itamarst:
The PR is buggy, `copy`'s source and destination are swapped... which raises the interesting question of why tests are passing :grinning:

itamarst:
Will try to push updated version later.

itamarst:
OK, hopefully fixed the bug in original diff.

calavera:
@carlosmn can you take a look at this? I just bumped into that issue with Go 1.6

carlosmn:
I don't get _how_ this fixes anythiing. We already had a `[]byte` so we're already dealing with a bunch of bytes without any interpretation. Copying them into another buffer of the same type clearly doesn't make it any safer, why does the runtime think it does?

carlosmn:
It turns out to be heavily dependent on how and what exactly you pass to the C function. CGo may or may not be able to know that we're talking about a slice and put the pointer check in the wrong place, checking for _all_ the contents of the slice, which obviously contains a pointer to the data it refers to. I've come up with an incantation that'll work without the need for a copy, so I'll send that as a PR with your test.

carlosmn:
#337 provides a fix for this issue which makes the checker happy without the need to copy a potentially large array.

carlosmn:
Fixed via #337 which is also included in `next`. Thanks for the test case.

itamarst:
Thanks for fixing this!

PR #307, opened 2016-04-30
Build on Windows using MSYS2

clns:
Similar to Unix, this allows running 'make install' from a MinGW shell
on Windows, installed using MSYS2.

Prerequisites:

- Install MSYS2 from http://msys2.github.io and follow the instructions
  on that page
- Install MinGW as explained here http://stackoverflow.com/a/30071634/1104534
- In addition to 'mingw-w64-x86_64-gcc', you'll also need the following:
  $ pacman -S make mingw-w64-x86_64-cmake mingw-w64-x86_64-pkg-config

Build libgit2 and git2go:

- Open the "MinGW-w64 Win64 Shell" (or 32-bit) and run 'make install'
from the root of git2go. 'git2go.a' should install in
'GOPATH\pkg\windows_amd64\github.com\libgit2'.

When building your app, you'll need gcc available in your PATH,
so you might want to add 'C:\msys64\mingw64\bin' to the PATH.

### DIFF
--- a/script/build-libgit2-static.sh
+++ b/script/build-libgit2-static.sh
@@ -5,15 +5,26 @@ set -ex
 VENDORED_PATH=vendor/libgit2
 
 cd $VENDORED_PATH &&
-mkdir -p install/lib &&
 mkdir -p build &&
-cd build &&
-cmake -DTHREADSAFE=ON \
+cd build
+
+if [ "$OSTYPE" = "msys" ]; then
+    cmake -DTHREADSAFE=ON \
       -DBUILD_CLAR=OFF \
       -DBUILD_SHARED_LIBS=OFF \
       -DCMAKE_C_FLAGS=-fPIC \
       -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
-      -DCMAKE_INSTALL_PREFIX=../install \
-      .. &&
+      -DCMAKE_INSTALL_PREFIX=install \
+      -G "MSYS Makefiles" \
+      ..
+else
+    cmake -DTHREADSAFE=ON \
+      -DBUILD_CLAR=OFF \
+      -DBUILD_SHARED_LIBS=OFF \
+      -DCMAKE_C_FLAGS=-fPIC \
+      -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
+      -DCMAKE_INSTALL_PREFIX=install \
+      ..
+fi
 
-cmake --build .
+cmake --build . --target install


ktsakas:
I get the following error:

```
\# github.com/libgit2/git2go
gcc: error: Files: No such file or directory
gcc: error: (x86)/libgit2/include: No such file or directory
```

I think the reason is that go build does no escape spaces on Windows paths properly.
I suspect the problem is in the build.go file.

torinaki:
You safe my day!

carlosmn:
The only difference between these seems to be the generator. I'd rather see that part be part of common code and then add the different args in whichever conditional is appropriate.

clns:
Indeed the only change is the generator, however it took me a while to figure out all the bits and pieces to be able to compile it on Windows. Also this PR includes the details on how to prepare the MSYS2 environment.

I would suggest we either use this change, or at least add the instructions somewhere in the README so it's easier for anyone to find it, and maybe change the file to accept a generator given as arg in cli.

mschenk42:
Hoping this might help others.  Took me a good amount of time to get this to work with the master branch.  I'm pretty happy with the process.  It is little kludgy for a Go build.  I wasn't able to vendor git2go like other go packages.

Take a look at the `appveyour.yml` and the `script` directory for the Windows build process in this project https://github.com/git-time-metric/gtm.

Fjolnir-Dvorak:
hmm, I get another error when trying to run `$ make install` on a fresh install of msys:
(Sorry, still need to try to get mingw to output english on a german computer)

```
[100%] Linking C static library libgit2.a
make[3]: Verzeichnis „/c/cygwin64/home/THORRAFNRIKR/workspace/go/src/github.com/libgit2/git2go/vendor/libgit2/build“ wird verlassen
[100%] Built target git2
make[2]: Verzeichnis „/c/cygwin64/home/THORRAFNRIKR/workspace/go/src/github.com/libgit2/git2go/vendor/libgit2/build“ wird verlassen
make[1]: Verzeichnis „/c/cygwin64/home/THORRAFNRIKR/workspace/go/src/github.com/libgit2/git2go/vendor/libgit2/build“ wird verlassen
go install ./...
# github.com/libgit2/git2go
.\blame.go:4:18: fatal error: git2.h: No such file or directory
compilation terminated.
make: *** [Makefile:8: install] Fehler 2
```

In short: It is successfully compiling a static libgit2.a binary and is not able to build the wrapper afterwards. The error code is 2.

mschenk42:
The windows build is working again for this project https://github.com/git-time-metric/gtm. Review the appveyor.yml, .travis.yml and scripts directory for how to.

PR #319, opened 2016-05-29
Remote Refinements

netnose:
- Fixed credentialsCallback return value for missing callback
- Added Remote Rename
- Added Remote Disconnect


remote.go
```diff
@@ -454,6 +454,25 @@ func (o *Remote) PushUrl() string {
...
+	defer C.free(unsafe.Pointer(cremote))
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ret := C.git_remote_rename(&cproblems, c.repo.ptr, cremote, cnewname)
+	if ret < 0 {
+		return MakeGitError(ret)
+	}
+	C.git_strarray_free(&cproblems)
```
carlosmn: We're just throwing away the issues we found here, which makes it impossible for a user of the bindings to tell the user about unexpected references they might find. We have to return this as well.

carlosmn:
Looks good other than throwing away the errors in rename.

netnose:
You're right, this signature looks better.

remote.go
```diff
@@ -454,6 +454,25 @@ func (o *Remote) PushUrl() string {
...
+	cremote := C.CString(remote)
+	defer C.free(unsafe.Pointer(cremote))
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ret := C.git_remote_rename(&cproblems, c.repo.ptr, cremote, cnewname)
+	if ret < 0 {
+		problems := makeStringsFromCStrings(cproblems.strings, int(cproblems.count))
+		return problems, MakeGitError(ret)
```
carlosmn: An error indicates an error in the function, but failure to rename some refspecs because they do not follow the standard pattern is not an error condition. The error array is returned when `ret == 0`. As the documentation for this function says, the ownership of the strings passes to the caller, so you need to free the string array.

netnose: The second line of the function frees the string array.

carlosmn:
Thanks.

PR #321, opened 2016-05-29
Checkout callbacks

netnose:
- Added CheckoutNotifyType mapping git_checkout_notify_t
- Added CheckoutOpts.NotifyFlags of type CheckoutNotifyType
- Added CheckoutNotifyCallback mapping git_checkout_notify_cb
- Added CheckoutProgressCallback mapping git_checkout_progress_cb

### DIFF
--- a/checkout.go
+++ b/checkout.go
@@ -96,6 +96,9 @@ func (opts *CheckoutOpts) toC() *C.git_checkout_options {
 
 //export checkoutNotifyCallback
 func checkoutNotifyCallback(why C.git_checkout_notify_t, cpath *C.char, cbaseline, ctarget, cworkdir, data unsafe.Pointer) int {
+	if data == nil {
+		return 0
+	}
 	path := C.GoString(cpath)
 	var baseline, target, workdir DiffFile
 	if cbaseline != nil {
@@ -116,6 +119,9 @@ func checkoutNotifyCallback(why C.git_checkout_notify_t, cpath *C.char, cbaselin
 
 //export checkoutProgressCallback
 func checkoutProgressCallback(path *C.char, completed_steps, total_steps C.size_t, data unsafe.Pointer) int {
+	if data == nil {
+		return 0
+	}
 	opts := pointerHandles.Get(data).(CheckoutOpts)
 	if opts.ProgressCallback == nil {
 		return 0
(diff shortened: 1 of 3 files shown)


wrapper.c
```diff
@@ -10,6 +10,12 @@ void _go_git_populate_remote_cb(git_clone_options *opts)
 	opts->remote_cb = (git_remote_create_cb)remoteCreateCallback;
 }
 
+void _go_git_populate_checkout_cb(git_checkout_options *opts)
```
carlosmn: This doesn't just populate the checkout, so it should reflect that in its name.

checkout.go
```diff
@@ -83,6 +142,16 @@ func populateCheckoutOpts(ptr *C.git_checkout_options, opts *CheckoutOpts) *C.gi
...
+	ptr.notify_flags = C.uint(opts.NotifyFlags)
+	if opts.NotifyCallback != nil || opts.ProgressCallback != nil {
+		C._go_git_populate_checkout_cb(ptr)
+	}
+	if opts.NotifyCallback != nil {
+		ptr.notify_payload = pointerHandles.Track(*opts)
+	}
+	if opts.ProgressCallback != nil {
+		ptr.progress_payload = pointerHandles.Track(*opts)
+	}
```
carlosmn: Why track the same pointer twice?

carlosmn: And aren't we asking `Track()` here to track the structure itself? We should be asking it to track the _pointer_ which would be `opts` itself.

checkout.go
```diff
@@ -70,6 +94,41 @@ func (opts *CheckoutOpts) toC() *C.git_checkout_options {
...
+		return 0
+	}
+	return int(opts.NotifyCallback(CheckoutNotifyType(why), path, baseline, target, workdir))
+}
+
+//export checkoutProgressCallback
+func checkoutProgressCallback(path *C.char, completed_steps, total_steps C.size_t, data unsafe.Pointer) int {
+	if data == nil {
+		return 0
+	}
```
carlosmn: When would we be registering the progress (or notify) callback but not set the payload pointer? Wouldn't hitting this mean we're silently hiding a bug in the bindings?

wrapper.c
netnose: What do you mean?
Something like _go_git_populate_checkout_cbs or _go_git_populate_checkout_notify_and_progress_cbs?

checkout.go
netnose: We'll always set the payload pointer but this function is called by external code and checking the arguments is always a good practice.

carlosmn: It's only called as specified by our code, and if we're not passed `data` we have no reason to suspect anything else that gets passed in is valid. Why not check `path`? We're already assuming it's the right struct later on so we should either accept that we're asserting here, or return an error indicating that something has gone very wrong somewhere.

wrapper.c
carlosmn: Nevermind, I'm not sure what I was thinking.

PR #322, opened 2016-06-22
Add NewCredSshKeyFromMemory to the credentials helpers.

calavera:
Allowing to use public and private keys from memory
without reading them from disk and without using an
ssh agent.

Signed-off-by: David Calavera <david.calavera@gmail.com>

### DIFF
--- a/credentials.go
+++ b/credentials.go
@@ -44,17 +44,35 @@ func NewCredUserpassPlaintext(username string, password string) (int, Cred) {
 	return int(ret), cred
 }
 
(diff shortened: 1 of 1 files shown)


calavera:
@carlosmn would you mind merging this? the api change is very straight-forward and libgit2 has supported that call for quite a while.

calavera:
@carlosmn would you mind merging this? the api change is very straight-forward and libgit2 has supported that call for quite a while.

Ehekatl:
why this still not been merged ? It's pretty useful feature. @carlosmn

PR #323, opened 2016-06-23
Add DiffStats String

Coderlane:
This adds a String() function to DiffStats which returns the output of git_diff_stats_to_buf.


carlosmn:
Thanks.

PR #331, opened 2016-08-04
Fix: Error conversion issue

commondream:
We were seeing an issue with conversion of errors, and it ends up an error was being converted to `GitError rather than `*GitError`.


carlosmn:
Thanks, but it looks like the whole check isn't necessary. We've already checked whether it's signaling the end of the iteration just above, so removing the block would be preferred.

commondream:
@carlosmn OK, I've removed that check completely now.

PR #332, opened 2016-08-07
Rebase wrapper

ezwiebel:
Hello,

I needed `rebase` services to be wrapped in `git2go` so I propose this PR.
Should I propose it on `next` branch instead of `master` ?

There's still some work though : 
- test what happens when a conflict arise during the rebase chain
- manage `git_rebase_options` and wrap `git_rebase_init_options(...)`
- wrap `git_rebase_open(...)`

I'll probably do the test for merge conflicts as I will need it quite soon.
Other services are of no interest for me at the moment, let me know if you'd like them wrapped as I'm into the work. Shouldn't be too long.

Let me know what you think. Cheers.

### DIFF
--- a/rebase.go
+++ b/rebase.go
@@ -33,7 +33,7 @@ type RebaseOperation struct {
 	Exec string
 }
 
-func rebaseOperationFromC(c *C.git_rebase_operation) *RebaseOperation {
+func newRebaseOperationFromC(c *C.git_rebase_operation) *RebaseOperation {
 	operation := &RebaseOperation{}
 	operation.Type = RebaseOperationType(c._type)
 	operation.ID = newOidFromC(&c.id)
@@ -50,15 +50,6 @@ type Rebase struct {
 	ptr *C.git_rebase
 }
 
-// Abort aborts a rebase that is currently in progress, resetting the repository and working directory to their state before rebase began.
-func (rebase *Rebase) Abort() error {
-	err := C.git_rebase_abort(rebase.ptr)
-	if err < 0 {
-		return MakeGitError(err)
-	}
-	return nil
-}
-
 //RebaseInit initializes a rebase operation to rebase the changes in branch relative to upstream onto another branch.
 func (r *Repository) RebaseInit(branch *AnnotatedCommit, upstream *AnnotatedCommit, onto *AnnotatedCommit, opts *RebaseOptions) (*Rebase, error) {
 	runtime.LockOSThread()
@@ -66,7 +57,7 @@ func (r *Repository) RebaseInit(branch *AnnotatedCommit, upstream *AnnotatedComm
 
 	//TODO : use real rebase_options
 	if opts != nil {
-		return nil, errors.New("RebaseOptions Not implemented yet")
+		return nil, errors.New("RebaseOptions Not implemented yet, use nil for default opts")
 	}
 
 	if branch == nil {
@@ -90,6 +81,23 @@ func (r *Repository) RebaseInit(branch *AnnotatedCommit, upstream *AnnotatedComm
 	return newRebaseFromC(ptr), nil
 }
 
+// OperationAt gets the rebase operation specified by the given index.
+func (rebase *Rebase) OperationAt(index uint) *RebaseOperation {
+	operation := C.git_rebase_operation_byindex(rebase.ptr, C.size_t(index))
+	return newRebaseOperationFromC(operation)
+}
+
+// CurrentOperationIndex gets the index of the rebase operation that is currently being applied.
+// If the first operation has not yet been applied then this returns -1 (C.GIT_REBASE_NO_OPERATION).
+func (rebase *Rebase) CurrentOperationIndex() int {
+	return int(C.git_rebase_operation_current(rebase.ptr))
+}
+
+// OperationCount gets the count of rebase operations that are to be applied.
+func (rebase *Rebase) OperationCount() uint {
+	return uint(C.git_rebase_operation_entrycount(rebase.ptr))
+}
+
 // Next performs the next rebase operation and returns the information about it.
 // If the operation is one that applies a patch (which is any operation except GIT_REBASE_OPERATION_EXEC)
 // then the patch will be applied and the index and working directory will be updated with the changes.
@@ -104,7 +112,7 @@ func (rebase *Rebase) Next() (*RebaseOperation, error) {
 		return nil, MakeGitError(err)
 	}
 
-	return rebaseOperationFromC(ptr), nil
+	return newRebaseOperationFromC(ptr), nil
 }
 
 // Commit commits the current patch.
@@ -148,9 +156,16 @@ func (rebase *Rebase) Finish() error {
 	return nil
 }
 
-// OperationCount gets the count of rebase operations that are to be applied.
-func (rebase *Rebase) OperationCount() uint {
-	return uint(C.git_rebase_operation_entrycount(rebase.ptr))
+// Abort aborts a rebase that is currently in progress, resetting the repository and working directory to their state before rebase began.
+func (rebase *Rebase) Abort() error {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	err := C.git_rebase_abort(rebase.ptr)
+	if err < 0 {
+		return MakeGitError(err)
+	}
+	return nil
 }
 
 //Free frees the Rebase object and underlying git_rebase C pointer.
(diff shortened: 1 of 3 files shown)


ezwiebel:
uh oh, there are probably things I don't understand in the test run.
it runs normally on my side. any clue?

rebase.go
```diff
@@ -0,0 +1,242 @@
...
+		checkout_options:  *ro.CheckoutOptions.toC(),
+	}
+}
+
+func rewriteNotesRefToC(ref string) *C.char {
+	if ref == "" {
+		return nil
+	}
+	return C.CString(ref)
+}
```
carlosmn: This looks like it should have a much more generic name, since what it does is unrelated to note rewriting and it's simply mapping an empty string and `NULL`.

rebase.go
```diff
@@ -0,0 +1,242 @@
...
+
+// Rebase object wrapper for C pointer
```
carlosmn: This describes the structure mechanically, but almost everything is a wrapper, if we're going to have a comment here, it should describe the object itself and/or what you can do with it.

rebase.go
```diff
@@ -0,0 +1,242 @@
...
+type Rebase struct {
+	ptr *C.git_rebase
+}
+
+//RebaseInit initializes a rebase operation to rebase the changes in branch relative to upstream onto another branch.
+func (r *Repository) RebaseInit(branch *AnnotatedCommit, upstream *AnnotatedCommit, onto *AnnotatedCommit, opts *RebaseOptions) (*Rebase, error) {
```
carlosmn: We have prior art of naming `git_repository_init()` as `InitRepository()` since the namespacing in C and Go is  bit different, so this should be called `InitRebase()`.

rebase.go
```diff
@@ -0,0 +1,242 @@
...
+	err := C.git_rebase_init(&ptr, r.ptr, branch.ptr, upstream.ptr, onto.ptr, opts.toC())
+	if err < 0 {
+		return nil, MakeGitError(err)
+	}
+
+	return newRebaseFromC(ptr), nil
+}
+
+//RebaseOpen opens an existing rebase that was previously started by either an invocation of git_rebase_init or by another client.
+func (r *Repository) RebaseOpen(opts *RebaseOptions) (*Rebase, error) {
```
carlosmn: Like with the init function, this should be called `OpenRebase()`. The documentation shouldn't reference the C function `git_rebase_init` but our own.

rebase.go
```diff
@@ -0,0 +1,242 @@
...
+func (rebase *Rebase) OperationAt(index uint) *RebaseOperation {
+	operation := C.git_rebase_operation_byindex(rebase.ptr, C.size_t(index))
+	return newRebaseOperationFromC(operation)
+}
+
+// CurrentOperationIndex gets the index of the rebase operation that is currently being applied.
+// If the first operation has not yet been applied then this returns -1 (C.GIT_REBASE_NO_OPERATION).
+func (rebase *Rebase) CurrentOperationIndex() int {
+	return int(C.git_rebase_operation_current(rebase.ptr))
+}
```
carlosmn: This return value does not match how Go reports errors. I would expect a `uint` to return the current operation and an `error` to return any errors.

rebase.go
```diff
@@ -0,0 +1,242 @@
...
+
+// OperationCount gets the count of rebase operations that are to be applied.
+func (rebase *Rebase) OperationCount() uint {
+	return uint(C.git_rebase_operation_entrycount(rebase.ptr))
+}
+
+// Next performs the next rebase operation and returns the information about it.
+// If the operation is one that applies a patch (which is any operation except GIT_REBASE_OPERATION_EXEC)
```
carlosmn: There is no `GIT_REBASE_OPERATION_EXEC` in git2go, it's `RebaseOperationExec`.

rebase.go
```diff
@@ -0,0 +1,242 @@
...
+	err := C.git_rebase_next(&ptr, rebase.ptr)
+	if err < 0 {
+		return nil, MakeGitError(err)
+	}
+
+	return newRebaseOperationFromC(ptr), nil
+}
+
+// Commit commits the current patch.
+// You must have resolved any conflicts that were introduced during the patch application from the git_rebase_next invocation.
```
carlosmn: There is no `git_rebase_next` in git2go, we have `Next()`.

rebase.go
```diff
@@ -0,0 +1,242 @@
...
+	authorSig, err := author.toC()
+	if err != nil {
+		return err
+	}
+	defer C.git_signature_free(authorSig)
+
+	committerSig, err := committer.toC()
+	if err != nil {
+		return err
+	}
```
carlosmn: This leaks the committer signature.

rebase.go
```diff
@@ -0,0 +1,242 @@
...
+	defer runtime.UnlockOSThread()
+
+	err := C.git_rebase_abort(rebase.ptr)
+	if err < 0 {
+		return MakeGitError(err)
+	}
+	return nil
+}
+
+//Free frees the Rebase object and underlying git_rebase C pointer.
```
carlosmn: Missing space. The pointer isn't public so let's not mention it. "unmanaged resources" covers whatever we decide to put in there.

rebase.go
```diff
@@ -0,0 +1,242 @@
...
+	// RebaseOperationFixup No commit will be cherry-picked.  The client should run the given command and (if successful) continue.
+	RebaseOperationFixup RebaseOperationType = C.GIT_REBASE_OPERATION_FIXUP
+	// RebaseOperationExec No commit will be cherry-picked.  The client should run the given command and (if successful) continue.
+	RebaseOperationExec RebaseOperationType = C.GIT_REBASE_OPERATION_EXEC
+)
+
+// RebaseOperation describes a single instruction/operation to be performed during the rebase.
+type RebaseOperation struct {
+	Type RebaseOperationType
+	ID   *Oid
```
carlosmn: The other public id fields are named `Id`.

rebase_test.go
```diff
@@ -0,0 +1,402 @@
...
+
+// Tests
+
+func TestDefaultRebaseOptions(t *testing.T) {
+	opts, err := DefaultRebaseOptions()
+	checkFatal(t, err)
+
+	if opts.Version != 1 {
+		t.Error("Expected opts Version to equal 1, got ", opts.Version)
+	}
```
carlosmn: Don't hard-code 1 here, the point of the version field is to increase it.

rebase.go
ezwiebel: Ok, I followed go convention suggested by my IDE plugin

rebase.go
ezwiebel: Done

rebase.go
ezwiebel: Done

rebase.go
ezwiebel: Done.
I also added `ErrClassRebase` in `git.go` 
Hope this is correct

rebase.go
ezwiebel: Done

rebase.go
ezwiebel: Oups!
Done

rebase.go
ezwiebel: Done

rebase_test.go
ezwiebel: Not sure how to handle this. It's a test that checks the default version returned by DefaultRebaseOptions. After a look into C code, I thought it was a `#define` or a constant that was used as default version number.

carlosmn: The version it needs to return in the version that is specified by the libgit2 headers in the `#define`. That's what the compiler is building and that's what it will use in order to figure out the size of the struct, which has to match what we're telling libgit2 we have.

carlosmn:
Note that the test build has flagged one of the calls as having invalid thread locking. We need that in order for the error handling to work.

rebase_test.go
ezwiebel: Thanks for that. However I'm not sure how I should do as the cgo feature seems to be forbidden in tests... I will just drop this very low value test I think.

apflieger:
Any update  on this ?
BTW, thanks for taking care of it.

ezwiebel:
Just waiting for validation.

apflieger:
I'm using this for a while now. Works well. 2 things :
* The branch is easily rebasable. I forked it and rebased it on top of v25 branch.
* I have a minor suggestion : https://github.com/apflieger/git2go/commit/79d6f6c18c61894fd459c6312db419e4776702c7
Since `Repository.InitRebase` takes `opts *RebaseOptions` as argument, I makes more sense that `DefaultRebaseOptions` returns `opts *RebaseOptions` instead of `opts RebaseOptions`

ezwiebel:
Fair enough, I can take care of it.
But I can't decide to accept the PR.

apflieger:
🎉 🎈

Issue #314, opened 2016-05-21
ODB/RefDB interface needs refactoring


milgner:
After some experimentation, I found that the interface to the ODB and RefDB parts of libgit2 [do not work as expected](http://stackoverflow.com/questions/36670761/go-conflicting-types-between-import-and-c-library) due to the way cgo handles scoping.

Since it is not possible for other packages to pass in C types, functions like `NewOdbBackendFromC` can be removed and the related functionality refactored. It seems to me like the only way to properly do this would be to enhance the `OdbBackend` type by all the functions included in its C counterpart, `git_odb_backend`.
Let me know if you think this is a desirable feature, I'll gladly try to implement it.

MagicalTux:
Actually this can be solved easily by having both functions accept a unsafe.Pointer type rather than the actual classes, then cast internally. This way NewOdbBackendFromC and NewRefdbBackendFromC both works and the difference in scope of C (which is weird, by the way) is not a problem.

carlosmn:
I've merged the PR due to the scoping issues, but it doesn't have to be either-or. As @MagicalTux mentions, passing the pointer around is still possibly, and for the use-case you mention in SO, it's still the right thing to do, a the backend is implemented in C (although that's just a detail) and what you have is a pointer to a backend.

Being able to implement a backend in Go is valuable regardless of wishing to pass in pointers from third-party backends.

milgner:
Sounds good, thanks for the enhancement!

PR #348, opened 2016-09-30
Make New*BackendFromC take unsafe.Pointer as argument

MagicalTux:
Offers a solution to #314 

The change is simple and allows usage of NewOdbBackendFromC and NewRefdbBackendFromC.


carlosmn:
I really dislike having void pointers, but it doesn't look like the compiler gives us another option here.

PR #349, opened 2016-10-11
It's 2016 :tada:

Razer6:


josharian:
Typically copyright years don't get updated from year to year.

PR #351, opened 2016-10-20
Implement git_index_remove_directory in index wrapper

ezwiebel:
Here is a simple PR to implement git_index_remove_directory service into index wrapper.


carlosmn:
Thanks.

PR #353, opened 2016-10-31
Update to libgit2 a051ee3

carlosmn:


carlosmn:
This needed a couple of runs with 1.4. I suspect some race condition in the parallel tests, but I've been unable to replicate locally.

PR #358, opened 2016-12-22
Add support for indexers and alternate odb packfiles

lhchavez:
This allows for implementations of git servers written in Go.


lhchavez:
PTAL

lhchavez:
Abandoning in favor of https://github.com/libgit2/git2go/pull/416, since that one goes against the correct branch.

PR #360, opened 2017-01-12
Set PKG_CONFIG_PATH

kljensen:
Set PKG_CONFIG_PATH so that libgit2.pc can be found when building statically linked against libgit2.

### DIFF
--- a/script/with-static.sh
+++ b/script/with-static.sh
@@ -4,6 +4,7 @@ set -ex
 
 export BUILD="$PWD/vendor/libgit2/build"
 export PCFILE="$BUILD/libgit2.pc"
+export PKG_CONFIG_PATH=$BUILD
 
 FLAGS=$(pkg-config --static --libs $PCFILE) || exit 1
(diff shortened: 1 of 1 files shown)


carlosmn:
This file shouldn't have been in master originally and now this moved into the Go source files after re-enabling static building on master, but I don't get why this would have been necessary either way.

We tell `pkg-config` the exact file we're looking for, so when would we need to also specify it as a pkg-config directory?

script/with-static.sh
```diff
@@ -4,6 +4,7 @@ set -ex
 
 export BUILD="$PWD/vendor/libgit2/build"
 export PCFILE="$BUILD/libgit2.pc"
+export PKG_CONFIG_PATH=$BUILD
```
dbolkensteyn: i believe you'll want to use the installed .pc file, not the one from the temporary build folder.

lhchavez:
IIUC, https://github.com/libgit2/git2go/blob/master/git_static.go should now handle this scenario correctly, as long as `-tags static` is passed to the `go` invocation.

Please reopen if that is not true.

PR #363, opened 2017-02-06
Implement Time() for Commit struct

rickharris:
Follows the pattern from calculating `Signature.Time`.

I'd ultimately like to use this in the v0.24 and v0.25 branches, but I'm not sure what the best first step for that is. If this change is approved, I'll be glad to see it through to those branches however the maintainer's would like!

### DIFF
--- a/commit.go
+++ b/commit.go
@@ -9,6 +9,7 @@ import "C"
 
 import (
 	"runtime"
+	"time"
 	"unsafe"
 )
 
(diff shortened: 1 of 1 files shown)


carlosmn:
This is just a convenience method to avoid going though the committer signature, so we should just be doing that instead of dropping into C ourselves. It exists separately partly just to avoid the overhead in C of dealing with the signature, but in Go w can write

```go
func (c Commit) Time() time.Time {
    return c.Committer().When
}
```

which reduces the duplication in conversion logic and gets us to the same information.

PR #372, opened 2017-04-14
[WIP] Use Go's TLS stream instead of whatever the library ships with

carlosmn:
This lets us have a more consistent set of features across platforms and avoids having to worry about OpenSSL's locking support as much (since _crypto_ itself doesn't seem to need it).

There is no support for proxies in this state. libgit2 itself only supports proxies either on Windows (at a HTTP level which wouldn't use this code anyway) and by piggy-backing off of curl, which this doesn't do either.

I'm not familiar with what proxy support exists in the Go ecosystem, but I assume there must be something that'll let us do HTTP tunneling.

### DIFF
--- a/git.go
+++ b/git.go
@@ -137,6 +137,10 @@ func init() {
 		panic("libgit2 was not built with threading support")
 	}
 
+	if err := RegisterManagedTls(); err != nil {
+		panic(err)
+	}
+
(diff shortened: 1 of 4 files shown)


lhchavez:
I just landed https://github.com/libgit2/git2go/pull/810, which was heavily based upon this change, with proxy support!

PR #374, opened 2017-04-17
[WIP] Use a managed HTTP(S) transport implementation

carlosmn:
This is a bit like #372 but goes further by replacing the whole HTTP stack and just leaving the logic in libgit2. This lets us use the Go HTTP code as well, instead of the one in the library.

The current state is enough to let the tests work but there is no custom proxy configuration available and we don't currently test that authentication actually works.

Proxy support depends on libgit2/libgit2#4206 or similar getting merged. I guess we'll have to add poxyproxy to our tests as well so it doesn't regress.

While this is useful by itself (by taking us completely away from OpenSSL and libcurl in libgit2), a potentially more exciting outcome might be to build on the lessons from this to build a managed ssh transport, which would eliminate the need for libssh2.

### DIFF
--- a/http.go
+++ b/http.go
@@ -208,6 +208,9 @@ func (self *ManagedHttpStream) sendRequest() error {
 		if resp.StatusCode == http.StatusUnauthorized {
 			resp.Body.Close()
 			var cred *C.git_cred
+
+			runtime.LockOSThread()
+			defer runtime.UnlockOSThread()
 			ret := C.git_transport_smart_credentials(&cred, self.owner.owner, nil, C.GIT_CREDTYPE_USERPASS_PLAINTEXT)
 
 			if ret != 0 {
--- a/http.go
+++ b/http.go
@@ -25,6 +25,8 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"reflect"
 	"runtime"
@@ -128,6 +130,7 @@ func (self *ManagedTransport) Action(url string, action SmartService) (SmartSubt
 }
 
 func (self *ManagedTransport) Close() error {
+	self.client = nil
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


AaronO:
@carlosmn This is pretty exciting, I actually hacked on implementing pure-go transports a few months ago. Would be great to reduce the surface area of C code and potential memory leaks.

We've had issues with `libgit2` memory leaks on long running daemons, the `go` heap is rarely over `10MB` but the process' resident memory grows to gigabytes.

I also explored a more radical approach and hacked together a fairly useable `git2go` compatible lib called `go2git` that wraps https://github.com/src-d/go-git (so it's 100% go with no C code).

http.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+
+	var p []byte
+	header := (*reflect.SliceHeader)(unsafe.Pointer(&p))
+	header.Cap = int(l)
+	header.Len = int(l)
+	header.Data = uintptr(unsafe.Pointer(data))
+
+	n, err := stream.Read(p)
+	if err != nil {
```
lhchavez: Per https://golang.org/pkg/io/#Reader, this should be

```suggestion
	if n == 0 && err != nil {
```

lhchavez:
you should also be able to add something along the lines of

```diff
diff --git a/script/build-libgit2-static.sh b/script/build-libgit2-static.sh
index 1568ece..fc82c79 100755
--- a/script/build-libgit2-static.sh
+++ b/script/build-libgit2-static.sh
@@ -14,6 +14,10 @@ cmake -DTHREADSAFE=ON \
       -DCMAKE_C_FLAGS=-fPIC \
       -DCMAKE_BUILD_TYPE="RelWithDebInfo" \
       -DCMAKE_INSTALL_PREFIX=../install \
+      -DCURL=OFF \
+      -DUSE_HTTPS=OFF \
+      -DUSE_EXT_HTTP_PARSER=OFF \
       .. &&
 
 cmake --build . --target install
```

lhchavez:
Hi! I did a small refactoring of this change and

a) split it between creating a Transport interface and implementing the HTTP/HTTPs transports
b) did what you mentioned in the first commit and did the SSH transport

with that, libgit2 can be compiled without openssh and openssl! with a bit more work, we can probably make it _fully_ static.

you can find the refactor here: https://github.com/lhchavez/git2go/commits/transports what would be the best course of action to move this change forward?

lhchavez:
I just landed https://github.com/libgit2/git2go/pull/810, which was heavily based upon this change, with proxy support!

PR #381, opened 2017-05-16
clone: add RepositoryCreateCallback bindings

### LINKED ISSUE #300: add submodule clone support
I wanted to initialize and clone submodules using git2go. I came to the conclusion that this is currently not supported because a [part of the clone API](https://github.com/libgit2/libgit2/blob/ec5a43b60ab0dfcb42db8101edfca34be3d62947/src/submodule.c#L1152-L1155) is not implemented in git2go.

Is there a roadmap for this particular feature?

toxeus:
Closes #300

### DIFF
--- a/clone.go
+++ b/clone.go
@@ -4,6 +4,7 @@ package git
 #include <git2.h>
 
 extern void _go_git_populate_remote_cb(git_clone_options *opts);
+extern void _go_git_populate_repository_cb(git_clone_options *opts);
 */
 import "C"
 import (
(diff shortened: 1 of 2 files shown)


toxeus:
@carlosmn anything missing here?

carlosmn:
This is not a function which exists in the library. Binding the callbacks is a fine addition, but the submodule clone is not something we would have in the bindings.

toxeus:
Alright. I'll refactor the PR accordingly.

toxeus:
This is getting ridiculous. I'll just use the `os/exec` package to call the `git` command which already supports all the features I need. I'm not surprised anymore that `go get` is not using this library to do its work.

PR #383, opened 2017-06-06
Repository.MergeHeads returns Oid of heads being merged

craigmj:
MergeHeads returns a slice of the Oid's from the git_repository_mergehead_foreach function call.


craigmj:
Will return to this - have somehow messed up!

PR #385, opened 2017-06-11
rebase: fix int overflows on GIT_REBASE_NO_OPERATION

zchee:
Fix int overflows on `GIT_REBASE_NO_OPERATION`.
It will fail on Go tip(and maybe 1.8), See https://travis-ci.org/libgit2/git2go/jobs/234362572#L391-L392.


carlosmn:
`C.GIT_REBASE_NO_OPERATION` is not en error code, so the error isn't the mismatched sizes but a mismatch in the kind of return values that the Go code provides vs what libgit2 is actually returning.

carlosmn:
Superseded by #396

Issue #356, opened 2016-12-03
Hang in Fetch after Clone while converting to mirror


josharian:
The following program hangs most times I run it (but not always):

```go
package main

import (
	"log"

	git "gopkg.in/libgit2/git2go.v24"
)

// mirror does a clone --mirror of url into dest.
func mirror(url, dest string) (*git.Repository, error) {
	log.Printf("starting clone of %s into %s", url, dest)
	repo, err := git.Clone(url, dest, &git.CloneOptions{Bare: true})
	if err != nil {
		log.Printf("clone of %s into %s failed: %v", url, dest, err)
		return nil, err
	}
	log.Printf("finished clone of %s into %s", url, dest)

	cfg, err := repo.Config()
	if err != nil {
		return nil, err
	}
	remotes, err := repo.Remotes.List()
	if err != nil || len(remotes) == 0 || remotes[0] != "origin" {
		return nil, err
	}

	log.Printf("converting remote origin to mirror")
	cfg.SetBool("remote.origin.mirror", true)
	cfg.SetString("remote.origin.fetch", "+refs/*:refs/*")
	remote, err := repo.Remotes.Lookup("origin")
	if err != nil {
		return nil, err
	}
	log.Printf("starting mirror fetch of %s into %s", url, dest)
	err = remote.Fetch(nil, &git.FetchOptions{}, "")
	if err != nil {
		log.Printf("mirror fetch of %s into %s failed: %v", url, dest, err)
		return nil, err
	}
	log.Printf("finished mirror fetch of %s into %s", url, dest)

	return repo, nil
}

func main() {
	dest := "/tmp/review_clone_go"
	url := "https://go.googlesource.com/review"
	_, err := mirror(url, dest)
	if err != nil {
		log.Printf("fail: %v", err)
		return
	}
	log.Print("success")
}
```

It hangs during the call to remote.Fetch. Sample run, killed manually by SIGQUIT when it hung:

```bash
$ go run bug.go 
2016/12/02 16:22:39 starting clone of https://go.googlesource.com/review into /tmp/review_clone_go
2016/12/02 16:22:41 finished clone of https://go.googlesource.com/review into /tmp/review_clone_go
2016/12/02 16:22:41 converting remote origin to mirror
2016/12/02 16:22:41 starting mirror fetch of https://go.googlesource.com/review into /tmp/review_clone_go
^\SIGQUIT: quit
PC=0x7fffc1f20f4e m=0
signal arrived during cgo execution

[...]

Issue #352, opened 2016-10-31
SEGFAULT on Repository.LookupBlob


AaronO:
### Stacktrace

```
runtime stack:
runtime.throw(0xbd91e0, 0x2a)
        /usr/local/go/src/runtime/panic.go:547 +0x90
runtime.sigpanic()
        /usr/local/go/src/runtime/sigpanic_unix.go:12 +0x5a

goroutine 2654459 [syscall, locked to thread]:
runtime.cgocall(0x8881fd, 0xc823f46eb8, 0xc800000000)
        /usr/local/go/src/runtime/cgocall.go:123 +0x11b fp=0xc823f46e68 sp=0xc823f46e38
github.com/libgit2/git2go._Cfunc_git_object_lookup(0xc8202b2048, 0x7fe296dffa20, 0xc8237b0760, 0x3, 0x0)
        ??:0 +0x41 fp=0xc823f46eb8 sp=0xc823f46e68
github.com/libgit2/git2go.(*Repository).lookupType(0xc822bc2980, 0xc8237b0760, 0x3, 0x0, 0x0, 0x0)
        /go/src/github.com/libgit2/git2go/repository.go:172 +0x288 fp=0xc823f46f60 sp=0xc823f46eb8
github.com/libgit2/git2go.(*Repository).LookupBlob(0xc822bc2980, 0xc8237b0760, 0xc8237b0760, 0x0, 0x0)
        /go/src/github.com/libgit2/git2go/repository.go:203 +0x44 fp=0xc823f46fa8 sp=0xc823f46f60
```

### Calling code

```
func GetBlob(repo *git.Repository, sha string) (*BlobResponse, error) {
	// Ensure SHA is OID
	oid, err := git.NewOid(sha)
	if err != nil {
		return nil, err
	}

	// Get blob
	blob, err := repo.LookupBlob(oid)
	if err != nil {
		return nil, err
	}

        // More code

        return nil, nil
}
```

### Environment

- `go` : `1.6`
- `git2go` : `37d3c2d9ad4c4e970cac02faec8ad184412c34e6`
- `libgit2` :  `97e57e8770132d61ff2c36bee2de2c7ac5c9d609`

carlosmn:
This might be another instance of #334, #373 and #356 where Go's GC is a bit too eager for the assumptions the code makes.

carlosmn:
This should be fixed via #393

Issue #334, opened 2016-08-20
segmentation violation on submodule update


dealloc:
I'm trying to add a submodule to an existing Git repository, using Git I'd do this

``` sh
git submodule add <url>
```

After searching through the documentation and scarce examples this is what I came up with

``` go
repo, err := git.OpenRepository(".")
if err != nil {
    panic(err)
}

sub, _ := repo.Submodules.Add("https://github.com/dealloc/ioc.git", "ioc", true)
sub.Init(true)
err = sub.Update(true, &git.SubmoduleUpdateOptions{
    &git.CheckoutOpts{},
    &git.FetchOptions{},
    git.CheckoutForce | git.CheckoutUpdateSubmodules,
})
fmt.Println("Submodule:", sub.Name())
```

If I run this code however the program panics

```
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x7f405eb0c728]

runtime stack:
runtime.throw(0x6606d0, 0x2a)
    /usr/lib/go/src/runtime/panic.go:566 +0x95
runtime.sigpanic()
    /usr/lib/go/src/runtime/sigpanic_unix.go:12 +0x2cc

goroutine 1 [syscall, locked to thread]:
runtime.cgocall(0x5dd520, 0xc42004b5f8, 0xc400000000)
    /usr/lib/go/src/runtime/cgocall.go:131 +0x110 fp=0xc42004b5b0 sp=0xc42004b570
github.com/dealloc/governor/vendor/gopkg.in/libgit2/git2go%2ev24._Cfunc_git_submodule_update(0x26acac0, 0xc400000001, 0xc420098c80, 0x0)
    ??:0 +0x4d fp=0xc42004b5f8 sp=0xc42004b5b0
github.com/dealloc/governor/vendor/gopkg.in/libgit2/git2go%2ev24.(*Submodule).Update(0xc42002a0b8, 0xc42007c501, 0xc42004b6e8, 0x0, 0x0)
    /home/dealloc/Documents/Projects/Personal/Go/src/github.com/dealloc/governor/vendor/gopkg.in/libgit2/git2go.v24/submodule.go:322 +0x26b fp=0xc42004b670 sp=0xc42004b5f8
main.addCommand(0xc4200b01b0)
    /home/dealloc/Documents/Projects/Personal/Go/src/github.com/dealloc/governor/add.go:19 +0x182 fp=0xc42004b730 sp=0xc42004b670
[...]

PR #393, opened 2017-07-08
KeepAlive all the things

carlosmn:
Newer Go runtimes' GC is particularly eager and doesn't even consider the receiver as a reference. Thus in order for structs wrapping the pointers not to be freed while in the middle of us using them in a function call, we must make sure that the wrapping struct is kept alive until after we return from.

This seems to be the cause of #352, #334 and #356 and this should fix the other places where this might happen.

### DIFF
--- a/blame.go
+++ b/blame.go
@@ -76,6 +76,7 @@ func (v *Repository) BlameFile(path string, opts *BlameOptions) (*Blame, error) 
 	defer runtime.UnlockOSThread()
 
 	ecode := C.git_blame_file(&blamePtr, v.ptr, cpath, copts)
+	runtime.KeepAlive(v)
 	if ecode < 0 {
 		return nil, MakeGitError(ecode)
 	}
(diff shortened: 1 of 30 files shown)


odeke-em:
@carlosmn thank you for working on this, for all the work and maintaining git2go!

I am checking in, having been paged in by @josharian who is currently on leave and away from his computer. By `Newer Go runtimes' GC` what would the older version be? I ask because the spread of versions in the closed issues is diverse, ranging from Go1.6 to Go1.8Beta1

Issue|GoVersion
---|---
https://github.com/libgit2/git2go/issues/352|Go1.6
https://github.com/libgit2/git2go/issues/334|Unspecified
https://github.com/libgit2/git2go/issues/356|Go1.7.4 and Go1.8Beta1

carlosmn:
AFAICT it was Go 1.6 which introduced the more eager garbage collector, but each version brings its own GC changes.

PR #400, opened 2017-07-17
Add binding for `git_index_add_frombuffer`

ramanenka:
It would be nice to have this binding.


carlosmn:
The blob's id and size are updated in the entry you pass into the C function, but this is thrown away in this implementation.

lhchavez:
> The blob's id and size are updated in the entry you pass into the C function, but this is thrown away in this implementation.

tried to add a test to ensure that the updated info gets propagated to the `entry`, but it turns out that this was never the case! the `source_entry` is never touched, and the `id` and `size` are only updated in the `entry` which is local to that function. So I sent out https://github.com/libgit2/libgit2/pull/5419 to fix the docs.

PR #402, opened 2017-07-22
clone_checkout_strategy removed

mikolajb:
There is a breaking API change in [libgit2 0.26](https://github.com/libgit2/libgit2/releases/tag/v0.26.0).

I guess `CloneCheckoutStrategy` also can be removed.

### DIFF
--- a/submodule.go
+++ b/submodule.go
@@ -15,7 +15,6 @@ import (
 type SubmoduleUpdateOptions struct {
 	*CheckoutOpts
 	*FetchOptions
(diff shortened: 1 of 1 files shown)


themalkolm:
Would be nice to merge it as some point to v26

sasso:
would it be possible to merge this to v26?

carlosmn:
Thanks, it looks like I made a bit of a mess with the v26 update. I fixed up a couple of things and merged your fix.

PR #404, opened 2017-08-14
align to libgit2 v0.26.0 api changes:

darekp:
`clone_checkout_strategy` has been removed from libgit2. The checkout strategy used to clone will
be the same strategy specified in `checkout_opts`


carlosmn:
Thanks, I merged #402 as it was submitted first.

PR #408, opened 2017-09-25
Added install step to static libgit2 build.

buddyspike:


carlosmn:
What problem are you trying to solve? We tell the linker to look for the static library in the build directory, so this would just copy it over to where nobody is looking for it.

buddyspike:
If you look at the generated `libgit2.pc` file in build directory, you will see that it's pointing the linker to lib and headers in install directory. this content is not produced if you don't run `install` target. To see this behaviour locally, you can `rm -rf build install` in libgit2 dir and try building again.

carlosmn:
We don't need those paths to be accurate, as we tell cgo and the linker where to find the headers and archive. We in fact point to the `libgit2.pc` in the build directory, and only need it on order to figure out which other libraries libgit2 ended up depending on.

I don't recall if older (c)go versions needed us to perform the install, but the ones we test against do let us specify `CFLAGS` and `LDFLAGS` so we know where to find it. At this point, the install location serves as a way to avoid accidentally installing this internal detail into system paths.

buddyspike:
Thanks @carlosmn. I will retract this PR. Static linking on master did not work for me for some reason without this step (on multiple machines with go 1.9). 
I found this while crawling https://github.com/libgit2/libgit2/blob/master/script/cibuild.sh#L24 ;-).

carlosmn:
Hm, it definitely should. This is exactly what we do on Travis and how I test locally so if it's not working, it should be worth investigating.

buddyspike:
Thanks again. Let me see if I can send you something reproducible.

PR #413, opened 2017-10-16
README: master wants to use install-static

carlosmn:


mhagger:
@carlosmn: Thanks. When building I also had a problem that `pkg-config` couldn't find `libgit2.pc`. I solved it by running

    PKG_CONFIG_PATH=vendor/libgit2/build make install-static

but there might be a better way.

carlosmn:
What Go version are you on? We pass the pc file directly to `pkg-config` in https://github.com/libgit2/git2go/blob/dc745f54c5690276bc3bdf83f27f751295a816ff/git_static.go#L9

/cc #408 which seems to be about the same issue

mhagger:
> What Go version are you on?

go1.8.1

mhagger:
Hmmm, maybe that error was happening before I figured out that I had to use `install-static` rather than `install`. Let me check again more carefully.

carlosmn:
It does give us an explanation, so we have that at least.

PR #416, opened 2017-12-05
Add support for indexers and alternate odb packfiles

lhchavez:
This allows for implementations of git servers written in Go.

### DIFF
--- a/indexer.go
+++ b/indexer.go
@@ -1,0 +1,99 @@
+package git
+
+/*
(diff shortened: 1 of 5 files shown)


lhchavez:
Now with tests! PTAL

lhchavez:
this change has been running in omegaup's production for more than a year with no issues. crossing fingers and merging.

PR #423, opened 2018-01-26
merge: add two missing AnnotatedCommit methods

josharian:
These should be reviewed somewhat skeptically; the implementation was pulled in monkey-see-monkey-do from similar calls.


merge.go
```diff
@@ -79,6 +88,22 @@ func (r *Repository) AnnotatedCommitFromRef(ref *Reference) (*AnnotatedCommit, e
...
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	var ptr *C.git_annotated_commit
+	ret := C.git_annotated_commit_from_revspec(&ptr, r.ptr, crevspec)
+	runtime.KeepAlive(r)
+	if ret < 0 {
+		return nil, MakeGitError(ret)
+	}
+	return newAnnotatedCommitFromC(ptr, r), nil
```
carlosmn: The `runtime.KeepAlive()` call needs to come after our last reference to what we need to keep alive. The annotated commit goes into `newAnnotatedCommitFromC` and we need to make sure `r` stays alive until we return from this function.

lhchavez: Fixed this and all other methods that have the same issue.

PR #425, opened 2018-01-26
merge: add missing MergeFileFlag constants

josharian:
While we're here, pull in comments as well.
While one can pop back and forth between godoc and libgit2 refs,
it's much nicer to have it in one place.

Note that MergeFileStyleSimplifyAlnum probably should have been called
merely MergeFileSimplifyAlnum (no "Style"). It's probably not worth
breaking backwards compatibility to fix it, but we avoid the mistake
going forwards.


carlosmn:
Thanks.

PR #429, opened 2018-02-07
cherrypick: wrap git_cherrypick_commit

josharian:
This PR comes with the same caveat as #423: It was implemented mainly by imitated nearby/similar code. Please review skeptically.

I think that CherryPickCommit would be better capitalization, but CherrypickCommit matches the capitalization of the existing Cherrypick method.


lhchavez:
would it be possible to add a test for this?

josharian:
Yes, although I'm not particularly inclined to return to it. I am no longer actively using git2go, and my experience contributing earlier was that PRs, issues, and questions got ignored a long time, so I turned my efforts elsewhere.

PR #430, opened 2018-02-09
tree: keep tree receiver alive as long as dependent entry is used

sprohaska:
I think `KeepAlive()` must be moved a bit down to protect `entry`.


carlosmn:
Yes, thankyou.

PR #431, opened 2018-02-15
reference: minor doc improvements

josharian:


carlosmn:
It does look like it's Go practice to refer to the nonsense receiver name. Thanks.

PR #432, opened 2018-02-15
git: simplify some Oid methods

josharian:


carlosmn:
How does this change the generated code? E.g. does `*oid == *oid2` put both values on the stack or does it know to reduce itself into the equivalent of `bytes.Equal`?

rmg:
I was curious about some of the implications of this as well so I took a look at what it generates..

_and hour later_

Both versions basically boil down to a call to calling an internal `runtime.memequal()` that is part of Go's runtime, however...

 * `*oid == *oid2` generats a direct call to `runtime.memequal()` using the address of each array
 * `bytes.Equal(oid[:], oid2[:])` allocates 2 new slices to pass to `bytes.Equal` which in turn calls `runtime.memequal()`

@carlosmn I think that's basically the opposite of what you were worried about, isn't it?

With that said, if the code actually defined temporary variables, then the compiler has to copy the bytes from each in to new arrays (because arrays are the underlying values while slices are merely references to some segment of an array).

...and now I'm going to go review all my own code to see where I can take advantage of this.

PR #440, opened 2018-04-18
Fix: Prevent overflow of unsigned options in git2

### LINKED ISSUE #439: DiffOptionsFlag causes an overflow in Windows
When building v27 branch on ~~Windows~~ for 32bit, I get this error:
```
vendor\github.com\libgit2\git2go\diff.go:441: constant 2147483648 overflows DiffOptionsFlag
```
I think the type of `DiffOptionsFlags` should be `uint` because in `git2` headers they are:
https://github.com/libgit2/libgit2/blob/master/include/git2/diff.h#L214

buddyspike:
Closes #439

### DIFF
--- a/diff.go
+++ b/diff.go
@@ -405,7 +405,7 @@ func (diff *Diff) Patch(deltaIndex int) (*Patch, error) {
 	return newPatchFromC(patchPtr), nil
 }
 
-type DiffOptionsFlag int
+type DiffOptionsFlag uint
 
 const (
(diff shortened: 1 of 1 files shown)


carlosmn:
Enum types in C are representable with an `int`, we can't choose to pretend it's an unsigned value. This would just muddy the waters wrt interoperability with the library.

libgit2/libgit2#4680 has fixed this meanwhile by making the enum be conformant to the C89 standard again and master now contains that fix.

PR #445, opened 2018-06-28
static: use pkg-config exclusively when using it

rmg:
When using the static linking option on platforms that use pkg-config,
use ONLY pkg-config to get the CFLAGS and LDFLAGS. This prevents pulling
in dependencies and flags for any non-vendored version that may be
present on the host.

The main practical effect of this is that if someone doesn't need/want
any sort of remote access support at all they can completely disable
libcurl, libssh2, libssl, etc and produce a smaller/simpler binary and
greatly simplify their build-time dependencies. When done properly, the
generated pkg-config file will tell cgo everything it needs to know.

This also prevents confusion if there is a system copy of libgit2 that
is being given priority over the vendored build.

### DIFF
--- a/script/build-libgit2-static.sh
+++ b/script/build-libgit2-static.sh
@@ -16,4 +16,4 @@ cmake -DTHREADSAFE=ON \
       -DCMAKE_INSTALL_PREFIX=../install \
       .. &&
 
-cmake --build .
+cmake --build . --target install
--- a/git_static.go
+++ b/git_static.go
@@ -3,9 +3,8 @@
 package git
 
 /*
-#cgo CFLAGS: -I${SRCDIR}/vendor/libgit2/include
-#cgo LDFLAGS: -L${SRCDIR}/vendor/libgit2/build/ -lgit2
-#cgo windows LDFLAGS: -lwinhttp
+#cgo windows CFLAGS: -I${SRCDIR}/vendor/libgit2/include
+#cgo windows LDFLAGS: -L${SRCDIR}/vendor/libgit2/build/ -lgit2 -lwinhttp
 #cgo !windows pkg-config: --static ${SRCDIR}/vendor/libgit2/build/libgit2.pc
 #include <git2.h>
 


rmg:
This patch was required to be able to do the following in my builds:
```
LIBGIT2 := vendor/gopkg.in/libgit2/git2go.v27/vendor/libgit2
LIBGIT2PC := $(LIBGIT2)/build/libgit2.pc

my_cmd: cmd/my_cmd/*.go $(LIBGIT2PC)
	go build -o $@ --tags "static" ./cmd/my_cmd

$(LIBGIT2)/CMakeLists.txt:
	git submodule update --init $(@D)

$(LIBGIT2PC): $(LIBGIT2)/CMakeLists.txt
	mkdir -p $(@D) $(LIBGIT2)/install/lib
	cd $(@D) && \
		cmake -DTHREADSAFE=ON \
			-DBUILD_CLAR=OFF \
			-DUSE_SSH=OFF \
			-DUSE_HTTPS=OFF \
			-DCURL=OFF \
			-DBUILD_SHARED_LIBS=OFF \
			-DCMAKE_C_FLAGS=-fPIC \
			-DCMAKE_INSTALL_PREFIX=../install \
			-DCMAKE_BUILD_TYPE="MinSizeRel" \
		.. && \
	  cmake --build . && \
	  make install
```

script/build-libgit2-static.sh
```diff
@@ -16,4 +16,5 @@ cmake -DTHREADSAFE=ON \
       -DCMAKE_INSTALL_PREFIX=../install \
       .. &&
 
-cmake --build .
+cmake --build . &&
```
lhchavez: You can use `cmake --build . --target install` as in https://github.com/libgit2/libgit2/blob/f4633791104fa5be75115fd4c7fab0c097da345e/script/cibuild.sh#L27

rmg:
@lhchavez thanks for the tip. I've incorporated in to the PR.

carlosmn:
The builds keep working, so it doesn't look like it breaks the existing linking types. Ta.

PR #446, opened 2018-07-03
Add revert functionality

### LINKED ISSUE #436: git revert
I don't see any go bindings for revert functionality in libgit. Is this intentional or something which is missing and needs to be added? I can create a patch, but wanted to check with contributors if there is a workaround or specific reason not to do it.

rgburke:
Closes #436


revert.go
```diff
@@ -0,0 +1,102 @@
...
+
+	var cOpts *C.git_revert_options
+
+	if revertOptions != nil {
+		cOpts = revertOptions.toC()
+		defer freeRevertOptions(cOpts)
+	}
+
+	ecode := C.git_revert(r.ptr, commit.cast_ptr, cOpts)
+	runtime.KeepAlive(commit)
```
lhchavez: oh missed something: you also need

```go
runtime.KeepAlive(r)
```

here and in `RevertCommit`.

rgburke: Thanks for the review, `runtime.KeepAlive(r)` has been added.

revert.go
```diff
@@ -0,0 +1,103 @@
...
+#include <git2.h>
+*/
+import "C"
+import (
+	"runtime"
+)
+
+// RevertOptions contains options for performing a revert
+type RevertOptions struct {
+	Version      uint
```
carlosmn: Version is not a user setting but decided at compile time to let libgit2 know what idea of the struct these bindings know about. It should always be set to `C.GIT_REVERT_OPTIONS_VERSION`. You use this for the default, but override for the case where someone passes in something. You should also prefer to use the function to initialize the default (though right now in this case it is admitedly not a complex default).

rgburke: Thanks for the feedback. `Version` has been removed from `RevertOptions`.  The value of `version` in the C struct is now always set to `C.GIT_REVERT_OPTIONS_VERSION`.

PR #447, opened 2018-07-03
git2go: small fixes to odb module

walkenzoy:
- Fix couple cgo issues in odb.Write() and odb.Hash(). This is the
  same issue - and same solution - as repo.CreateBlobFromBuffer()
  used to have.

- Add test for odb.Read()

### DIFF
--- a/odb.go
+++ b/odb.go
@@ -80,15 +80,19 @@ func (v *Odb) Exists(oid *Oid) bool {
 
 func (v *Odb) Write(data []byte, otype ObjectType) (oid *Oid, err error) {
 	oid = new(Oid)
-	var cptr unsafe.Pointer
-	if len(data) > 0 {
-		cptr = unsafe.Pointer(&data[0])
-	}
 
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
-	ret := C.git_odb_write(oid.toC(), v.ptr, cptr, C.size_t(len(data)), C.git_otype(otype))
+	var size C.size_t
+	if len(data) > 0 {
+		size = C.size_t(len(data))
+	} else {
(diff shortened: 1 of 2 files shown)


walkenzoy:
I want to add - OdbReadStream.Read() and OdbWriteStream.Write() also look suspicious to me, but I wasn't able to write a test that triggers any issue that might be there...

walkenzoy:
Ping - is there anything else I can do to unblock this ?

carlosmn:
What is the problem that `CreateBlobFromBuffer` used to have that this is meant to be fixing?

walkenzoy:
Hey Carlos,

CreateBlobFromBuffer() used to panic due to cgo issues when being fed bytes from a short bytes.Buffer{} - IIRC, bytes.Buffer{} allocates the first 64 bytes in place, and when cgo sees a pointer to these first 64 bytes, it thinks it is passed a pointer to a go object (the bytes.Buffer) that holds pointers to other go objects (the first 64 bytes) and panics.

The same issue is present in odb.Write() and odb.Hash(). The updated TestOdbHash() will trigger this cgo panic by using doublePointerBytes(), which uses the same construct as bytes.Buffer{}. The updated odb.Write() and odb.Hash() functions will work again without triggering the cgo issue anymore.

The changes to TestOdbRead() are not 100% necessary, but since I changed odb.Write() I wanted to increase its test coverage while I was there.

I also had cgo maintainer Ian Taylor check the changes for me at work - I would not normally resort to this kind of argument by authority, but low-level cgo stuff is obscure enough that I wouldn't be as confident about the changes if he hadn't, so probably worth adding here.

carlosmn:
I see, tanks for the explanation.

PR #448, opened 2018-07-07
Add support for mempack

lhchavez:
This allows the creation of git packfiles without needing to write them
to the filesystem.

### DIFF
--- a/mempack.go
+++ b/mempack.go
@@ -34,6 +34,7 @@ func NewMempack(odb *Odb) (mempack *Mempack, err error) {
 	}
 
 	ret = C.git_odb_add_backend(odb.ptr, mempack.ptr, C.int(999))
+	runtime.KeepAlive(odb)
 	if ret < 0 {
 		// Since git_odb_add_alternate() takes ownership of the ODB backend, the
 		// only case in which we free the mempack's memory is if it fails to be
--- a/mempack.go
+++ b/mempack.go
@@ -1,0 +1,97 @@
+package git
+
+/*
+#include <git2.h>
+#include <git2/sys/mempack.h>
+
+extern int git_mempack_new(git_odb_backend **out);
+extern int git_mempack_dump(git_buf *pack, git_repository *repo, git_odb_backend *backend);
+extern void git_mempack_reset(git_odb_backend *backend);
+extern void _go_git_odb_backend_free(git_odb_backend *backend);
+*/
+import "C"
+
+import (
+	"runtime"
+	"unsafe"
+)
+
+// Mempack is a custom ODB backend that permits packing object in-memory.
+type Mempack struct {
+	ptr *C.git_odb_backend
+}
+
+// NewMempack creates a new mempack instance and registers it to the ODB.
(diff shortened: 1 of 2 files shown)


mempack.go
```diff
@@ -0,0 +1,97 @@
...
+func (mempack *Mempack) Reset() {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	C.git_mempack_reset(mempack.ptr)
+}
+
+// Free frees the mempack and its resources.
+func (mempack *Mempack) Free() {
+	runtime.SetFinalizer(mempack, nil)
```
carlosmn: This still needs to actually free the mempack, not just unset the finalizer.

mempack.go
```diff
@@ -0,0 +1,97 @@
...
+	return C.GoBytes(unsafe.Pointer(buf.ptr), C.int(buf.size)), nil
+}
+
+// Reset resets the memory packer by clearing all the queued objects.
+//
+// This assumes that Mempack.Dump has been called before to store all the
+// queued objects into a single packfile.
+func (mempack *Mempack) Reset() {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
```
carlosmn: We only need to lock the thread if we're going to deal with errors.

mempack.go
```diff
@@ -0,0 +1,97 @@
...
+
+	var repoPtr *C.git_repository
+	if repository != nil {
+		repoPtr = repository.ptr
+	}
+
+	ret := C.git_mempack_dump(&buf, repoPtr, mempack.ptr)
+	if ret < 0 {
+		return nil, MakeGitError(ret)
+	}
```
carlosmn: We need to keep `repository` alive until here, or the runtime might free it.

mempack.go
```diff
@@ -0,0 +1,97 @@
...
+	runtime.SetFinalizer(mempack, (*Mempack).Free)
+
+	ret = C.git_odb_add_backend(odb.ptr, mempack.ptr, C.int(999))
+	if ret < 0 {
+		// Since git_odb_add_alternate() takes ownership of the ODB backend, the
+		// only case in which we free the mempack's memory is if it fails to be
+		// added to the ODB. Mempack.Free() is actually just a no-op.
+		C._go_git_odb_backend_free(mempack.ptr)
+		mempack.Free()
```
carlosmn: The free should happen inside `Mempack.Free`.

carlosmn:
Actually since ownership transfers to the repository and the freeing only happens in the constructor, there is no need for a finalizer or `Mempack.Free()` at all.

mempack.go
lhchavez: Removed

mempack.go
lhchavez: Done

mempack.go
lhchavez: Removed

mempack.go
lhchavez: Removed.

carlosmn:
Thanks, I added an extra keepalive that was missing (these are _so_ annoying).

PR #449, opened 2018-07-07
Use git_buf_dispose() instead of git_buf_free()

lhchavez:
git_buf_free() is now deprecated. This change uses the new
git_buf_dispose(). It also updates `vendor/libgit2` to the latest
commit.


lhchavez:
fixed by 538a05d55c6051371dd1749af75a7b2e2d623d53

Issue #461, opened 2018-10-11
tree.Walk() is missing a context for callbacks


missionsix:
The v26 API for libgit2 specifies a `payload` context to be passed to the callback method:

From tree.h:
```c
/** Callback for the tree traversal method */
typedef int (*git_treewalk_cb)(
const char *root, const git_tree_entry *entry, void *payload);

/*
 * @param tree The tree to walk
 * @param mode Traversal mode (pre or post-order)
 * @param callback Function to call on each tree entry
 * @param payload Opaque pointer to be passed on each callback
 * @return 0 or an error code
 */
GIT_EXTERN(int) git_tree_walk(
	const git_tree *tree,
	git_treewalk_mode mode,
	git_treewalk_cb callback,
void *payload);
```

The `tree.Walk` call omits the opaque context:
```go
type TreeWalkCallback func(string, *TreeEntry) int

func (t Tree) Walk(callback TreeWalkCallback) error
```

Which makes it extremely difficult to recursively walk trees.

missionsix:
Propsal, use the empty interface:

```go
type TreeWalkCallback func(string, *TreeEntry, interface{}) int

func (t Tree) Walk(callback TreeWalkCallback, interface{}) error
```

carlosmn:
libgit2 has a payload because that's how you capture the context that you need for closure-like behaviour in C. In Go you can access whatever external context you need within the function you pass as a callback. This is the same way as everything else that takes a callback in Go (and many other languages) behave.

missionsix:
I understand, but this style forces your callback to become a closure over some scoped variables.

Using a context, similar to C style, allows for scope isolation.   I'm not against that style for some things, but with a callback that has to do a lot of work, it's better to move it out to a separate function, where you loose the closure.

lhchavez:
what's wrong with

```go
type explicitClosure struct {
   message string
}
func (c *explicitClosure) Walk (string, *TreeEntry) int {
   fmt.Printf("Hello, %s!\n", c.message)
   return 0
}

...

c := &explicitClosure{
  message: "world",
}
tree.Walk(c.Walk)
```

carlosmn:
This is also exactly how the Go stdlib handles this https://golang.org/pkg/path/filepath/#Walk though I'm not particularly a fan of these constructions either way.

PR #462, opened 2018-10-12
Fix Issue #461

missionsix:
Pass a callback context down to `git_tree_walk` instead of the
    callback pointer directly. This allows the `Walk` callback to
    more closely match that of libgit2 by attaching a callback
    payload.

    Additionally, add a unittest for tree.Walk() which verifies
    added tree entries and that the context data is valid.

### DIFF
--- a/tree.go
+++ b/tree.go
@@ -112,25 +112,36 @@ func (t Tree) EntryCount() uint64 {
 	return uint64(num)
 }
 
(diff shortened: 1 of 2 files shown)


carlosmn:
As mentioned in #461 the way in Go to get context is to use a closure and it's also how Go itself expects you to call it so this would move us away from the expectations in Go.

PR #463, opened 2018-10-25
Add index.Clear() to clear the index object

Nivl:
Simple PR that adds `index.Clear()`, which does nothing more than calling [`git_index_clear`](https://github.com/libgit2/libgit2/blob/bea65980c7a42e34edfafbdc40b199ba7b2a564e/include/git2/index.h#L404).

### DIFF
--- a/index.go
+++ b/index.go
@@ -145,6 +145,20 @@ func (v *Index) Path() string {
 	return ret
 }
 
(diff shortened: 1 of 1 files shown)


lhchavez:
would it be possible to add a test for this?

carlosmn:
It's just wrapping the call the most straightforward way so it's fine without a test. Thanks.

PR #466, opened 2018-11-15
Add support for CreateCommitFromIds

lhchavez:
This change adds support for CreateCommitFromIds from libgit2.

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -412,12 +412,19 @@ func (v *Repository) CreateCommitFromIds(
 
 	nparents := len(parents)
 	if nparents > 0 {
-		parentsarg = (**C.git_oid)(C.malloc(C.size_t(unsafe.Sizeof(uintptr(0)) * uintptr(nparents))))
+		// All this awful pointer arithmetic is needed to avoid passing a Go
+		// pointer to Go pointer into C. Other methods (like CreateCommits) are
+		// fine without this workaround because they are just passing Go pointers
+		// to C pointers, but arrays-of-pointers-to-git_oid are a bit special since
+		// both the array and the objects are allocated from Go.
+		var emptyOidPtr *C.git_oid
+		sizeofOidPtr := unsafe.Sizeof(emptyOidPtr)
+		parentsarg = (**C.git_oid)(C.calloc(C.size_t(uintptr(nparents)), C.size_t(sizeofOidPtr)))
 		defer C.free(unsafe.Pointer(parentsarg))
 		parentsptr := uintptr(unsafe.Pointer(parentsarg))
 		for _, v := range parents {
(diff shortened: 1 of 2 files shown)


repository.go
```diff
@@ -389,6 +391,66 @@ func (v *Repository) CreateCommit(
...
+	}
+
+	cmsg := C.CString(message)
+	defer C.free(unsafe.Pointer(cmsg))
+
+	var parentsarg **C.git_oid = nil
+
+	nparents := len(parents)
+	if nparents > 0 {
+		parentsarg = (**C.git_oid)(C.malloc(C.size_t(unsafe.Sizeof(uintptr(0)) * uintptr(nparents))))
```
carlosmn: It looks like this should be `C.calloc`. Should we be using sizeof on `*C.git_oid` rather than a generic pointer?

repository.go
```diff
@@ -389,6 +391,66 @@ func (v *Repository) CreateCommit(
...
+		defer C.free(unsafe.Pointer(parentsarg))
+		parentsptr := uintptr(unsafe.Pointer(parentsarg))
+		for _, v := range parents {
+			*(**C.git_oid)(unsafe.Pointer(parentsptr)) = v.toC()
+			parentsptr += unsafe.Sizeof(uintptr(0))
```
carlosmn: Can we avoid this pointer arithmetic? Can we generate the array in Go and pass that pointer to `C.git_commit_create_from_ids`?

lhchavez: IIRC this was a workaround for the dreaded [cgo argument has Go pointer to Go pointer](https://stackoverflow.com/questions/51483032/passing-go-pointer-to-cgo). admittedly, i didn't invest too much time trying more options, so let me give it another try.

repository.go
lhchavez: I chose malloc since i was writing to the entire array, but `C.calloc` looks a bit cleaner.

repository.go
lhchavez: Tried different things, but now I'm pretty sure we cannot avoid this :( added a comment.

repository.go
```diff
@@ -389,6 +391,73 @@ func (v *Repository) CreateCommit(
...
+	defer runtime.UnlockOSThread()
+
+	ret := C.git_commit_create_from_ids(
+		oid.toC(), v.ptr, cref,
+		authorSig, committerSig,
+		nil, cmsg, tree.toC(), C.size_t(nparents), parentsarg)
+
+	runtime.KeepAlive(v)
+	runtime.KeepAlive(oid)
+	runtime.KeepAlive(parents)
```
carlosmn: We also need to keep `tree` alive.

lhchavez: Done

Issue #256, opened 2015-09-21
Unused parameter in OpenOndisk(parent *Config, path string)


TheDahv:
This is _very_ minor and low priority, but something I noticed as I was working on a personal project.

It seems as though `parent` has no bearing on the result of the function. Am I misinterpreting its intent or usage, or is this just fodder for code cleanup?

https://github.com/libgit2/git2go/blob/master/config.go#L329

carlosmn:
It's quite possible this was a copy-paste error from `OpenLevel()` or a similar function. There is no parent here, since it's creating a new configuration object.

praveentiru:
I would like to work  on this. But, I am bit confused.
The issue refers OpenOndisk which now at https://github.com/libgit2/git2go/blob/master/config.go#L346 rather than https://github.com/libgit2/git2go/blob/master/config.go#L329 where we have OpenLevel.

I will submit appropriate patch assuming that we want to clean-up OpenOndisk.

carlosmn:
Fixed via #469

PR #469, opened 2018-12-25
Remove unused parameter in OpenOndisk #256

praveentiru:
This is to fix issue #256. Removing parent from OpenOndisk from method signature which is not used.

**Risk:** Change in method signature will impact dependent programs so, there is need to plan how to incorporate this change.


carlosmn:
Thanks, I think originally this was supposed to go the other way around and use the passed in configuration instead of the one we instantiate but it's been long enough like this that it's probably not happening (and that probably would be better as a method).

PR #470, opened 2018-12-28
Return io.EOF on OdbReadStream.Read()

lhchavez:
This change makes OdbReadStream.Read() comply with the usual io.Reader
semantics.


carlosmn:
Makes sense, thanks.

PR #477, opened 2019-01-06
Add support for Go 1.11 modules

lhchavez:
This change adds a `go.mod` file. An empty file is sufficient since this
project has no external dependencies. For people that want to use the static
version of libgit2, this module can be vendored and the following can be added
to their `go.mod` file:

    replace github.com/libgit2/git2go => ./vendor/github.com/libgit2/git2go

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,1 @@
+module github.com/libgit2/git2go


carlosmn:
Can you not use a project with go modules if the dependencies themselves aren't set up for it?

lhchavez:
you can, if you want to use the dynamically-linked version. what you can't do is use the vendoring trick to be able to build the static library, since using a vendored module requires the module to be explicitly a module.

carlosmn:
Fair enough I guess.

PR #478, opened 2019-01-16
Fix the Cred interface

lhchavez:
This change adds Cred.Free() and finalizers to prevent memory leaks. It
also makes the interface for Cred more idiomatic and return actual
errors intead of ints.

### DIFF
--- a/credentials.go
+++ b/credentials.go
@@ -7,7 +7,10 @@ package git
 git_credtype_t _go_git_cred_credtype(git_cred *cred);
 */
 import "C"
-import "unsafe"
+import (
+	"runtime"
+	"unsafe"
+)
 
 type CredType uint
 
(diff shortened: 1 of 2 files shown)


credentials.go
```diff
@@ -30,24 +39,41 @@ func (o *Cred) Type() CredType {
 	return (CredType)(o.ptr.credtype)
 }
 
-func credFromC(ptr *C.git_cred) *Cred {
-	return &Cred{ptr}
+func (o *Cred) Free() {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
```
carlosmn: There's no need to lock the thread here.

credentials.go
```diff
@@ -30,24 +39,41 @@ func (o *Cred) Type() CredType {
...
+
+	if o.ptr != nil {
+		C.git_cred_free(o.ptr)
+	}
```
carlosmn: There's no need to check for `nil` here.

carlosmn:
These types tie into the `CredentialsCallback` type. Unless `credentialsCallback` learns how to deal with this, code is going to break.

lhchavez:
> These types tie into the `CredentialsCallback` type. Unless `credentialsCallback` learns how to deal with this, code is going to break.

Done. We might need to update all the other callbacks, then.

PR #482, opened 2019-02-11
Make the static build script more portable

lhchavez:
This change stops using a relative cd (which as it turns out, _some_
shells don't like since `$0` is a file). Instead, this now uses
`dirname` and `git rev-parse --show-toplevel`, which should work
in more environments.

### DIFF
--- a/script/build-libgit2-static.sh
+++ b/script/build-libgit2-static.sh
@@ -2,7 +2,7 @@
 
 set -ex
 
-ROOT="$(cd "$0/../.." && echo "${PWD}")"
+ROOT="$(cd -- "$(dirname -- "$0")" && git rev-parse --show-toplevel)"
 BUILD_PATH="${ROOT}/static-build"
 VENDORED_PATH="${ROOT}/vendor/libgit2"
 


script/build-libgit2-static.sh
```diff
@@ -2,7 +2,7 @@
 
 set -ex
 
-ROOT="$(cd "$0/../.." && echo "${PWD}")"
+ROOT="$(cd -- "$(dirname -- "$0")" && git rev-parse --show-toplevel)"
```
jonEbird: I like the `echo` more than the `git rev-parse...` in case the code is ever without the git information, say from a git archive, e.g.

jonEbird:
I didn't realize this existed with I created #503

script/build-libgit2-static.sh
lhchavez: fair enough. as long as there is _a_ solution. i approved the other PR, let me see if we can get that one merged.

lhchavez:
no need for this anymore since #503 was merged!

PR #483, opened 2019-02-16
Fix receivers to Blob and Tree methods

lhchavez:
This change changes the signature of Blob and Tree methods so that they
have consistent receiver names and use pointers. This increases
consistency and avoids unnecessarily copying structs.


dbolkensteyn:
Relates to #513 and its PR #527

lhchavez:
no longer needed, thanks to #527

PR #484, opened 2019-02-16
Free() the copies of repository.LookupXxx()

lhchavez:
`repository.LookupXxx()` allocate new go `Object`s that have a reference
to a `C.git_object`. Those are then duplicated with `git_object_dup()`,
so the original `Object`s linger unnecessarily until the Go GC kicks in.

This change explicitly calls `Free()` on the originals to avoid
unnecessary accumulation of garbage.

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -194,6 +194,7 @@ func (v *Repository) LookupTree(id *Oid) (*Tree, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer obj.Free()
 
 	return obj.AsTree()
 }
@@ -203,6 +204,7 @@ func (v *Repository) LookupCommit(id *Oid) (*Commit, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer obj.Free()
 
 	return obj.AsCommit()
 }
@@ -212,6 +214,7 @@ func (v *Repository) LookupBlob(id *Oid) (*Blob, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer obj.Free()
 
 	return obj.AsBlob()
 }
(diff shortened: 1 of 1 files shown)


dbolkensteyn:
`allocObject()` is not actually duplicating the underlying `libgit2` object, it merely allocates a new Go object that keeps a reference to the same pointer: https://github.com/libgit2/git2go/blob/8766f9f36cfb30934b7133365aa3d1e91086bafa/object.go#L231

`AsBlob()` however invokes `dupObject`, which in turns calls `git_object_dup`, making this PR useful.

I am wondering: Why the object is being duplicated in the first place?

lhchavez:
> `allocObject()` is not actually duplicating the underlying `libgit2` object, it merely allocates a new Go object that keeps a reference to the same pointer:
> 
> https://github.com/libgit2/git2go/blob/8766f9f36cfb30934b7133365aa3d1e91086bafa/object.go#L231
> 
> `AsBlob()` however invokes `dupObject`, which in turns calls `git_object_dup`, making this PR useful.
> 
> I am wondering: Why the object is being duplicated in the first place?

thanks! updated the commit description to indicate the call to `git_object_dup`, making it more obvious where the dangling reference comes from.

lhchavez:
> I am wondering: Why the object is being duplicated in the first place?

The object is duplicated to avoid the case where one calls `t, _ := o.AsBlob()` and then the original `o` enters an invalid state if `t.Free()` is called.

I guess that since the whole lifetime of those objects are fully controlled in library code, doing some surgery in the original `Object`, grabbing its underlying `git_object` reference and setting it to `nil` could theoretically be possible (to avoid double-freeing it), but I wanted to avoid changing too much at once.

PR #486, opened 2019-02-20
Fix the static linking build script

takuji:
When I run `make install-static` on my mac, it fails with an error message below:
```
$ make install-static
./script/build-libgit2-static.sh
++ cd ./script/build-libgit2-static.sh/../..
./script/build-libgit2-static.sh: line 5: cd: ./script/build-libgit2-static.sh/../..: Not a directory
+ ROOT=
make: *** [build-libgit2] Error 1
```
This PR fixes the problem by replacing "$0/../.." with "$(dirname $0)/..".

Thanks.

### DIFF
--- a/script/build-libgit2-static.sh
+++ b/script/build-libgit2-static.sh
@@ -2,7 +2,7 @@
 
 set -ex
 
-ROOT="$(cd "$0/../.." && echo "${PWD}")"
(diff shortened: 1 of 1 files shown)


lhchavez:
This is basically https://github.com/libgit2/git2go/pull/482/files

lhchavez:
Thanks for the patch! We merged a functionally identical PR earlier (#503), so closing this since it would conflict.

PR #488, opened 2019-02-22
Transport plugins (v27)

brynbellomy:
I've added bindings for `git_register_transport(...)` and the `git_transport` interface.  I'm new to this library (and also new to C/Go interop), so I'd appreciate any feedback.


lhchavez:
Added support for this in https://github.com/libgit2/git2go/pull/809

PR #515, opened 2019-06-19
fix invalid guard forcing v27 instead of v28

Nivl:
A few months ago a PR has been made to [support v28](https://github.com/libgit2/git2go/commit/5fda6dd90191b1c51a1785ad7cabd2fd5b05e802), but the guard around the dynamic lib hasn't been updated. This PR fixes that.

### DIFF
--- a/git_dynamic.go
+++ b/git_dynamic.go
@@ -6,8 +6,8 @@ package git
 #include <git2.h>
 #cgo pkg-config: libgit2
 
-#if LIBGIT2_VER_MAJOR != 0 || LIBGIT2_VER_MINOR != 27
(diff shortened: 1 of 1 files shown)


rssathe:
@Nivl can we get this merged?

Nivl:
@rssathe Unfortunately that's not my call. Ping @carlosmn

cmbernard333:
Bumping this. Definitely want support for latest libgit2 with this library.

segevfiner:
Getting kinda silly. There is a v28 branch building with 0.28 statically but the guard there forces v27 which probably doesn't build.

PR #525, opened 2019-09-02
Raise the libgit version in dynamic building

Muhammad-Adel:
git lib version was raised only in the static build. This is raising in the dynamic build

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,1 +1,1 @@
-module github.com/libgit2/git2go
+module github.com/Muhammad-Adel/git2go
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1 +1 @@
-module github.com/libgit2/git2go
+module github.com/Muhammad-Adel/git2go
```
darix: I think that change is wrong.

lhchavez:
Thanks for the patch!

We merged a functionally identical PR earlier (#515) so if you pull, you should get the updated fixed version!

PR #531, opened 2020-01-18
Switch over the version constraints to v0.28

suhaibmujahid:


hackworks:
Request merging this change, it is blocking me from upgrading to libgit2 0.28 across Python and Go

lhchavez:
Thanks for the patch!

an identical PR was merged earlier today (#515), so closing this one since it would conflict with HEAD

suhaibmujahid:
@lhchavez thank you for looking into this pull request.

I want to let you know that this fix still not merged to the `v28` branch. Yes ((#515) have the same fix, but it was merged to the `master`.

lhchavez:
> @lhchavez thank you for looking into this pull request.
> 
> I want to let you know that this fix still not merged to the `v28` branch. Yes ((#515) have the same fix, but it was merged to the `master`.

branches are hard -_-. approved, will merge as soon as Travis says we are good to go.

QuLogic:
There really should be a new tag with this fix, since the last one cannot build as is.

lhchavez:
> There really should be a new tag with this fix, since the last one cannot build as is.

[v0.28.4.1](https://github.com/libgit2/git2go/releases/tag/v0.28.4.1) is now live!

Issue #533, opened 2020-02-03
git_mempack_reset returns int after libgit2 update


lydza:
Hello, all. I ran into an issue while running tests using git2go's master branch with libgit2's master branch. The fix is a quick one-liner, and everything ran smoothly after I made this change.

I tried to create a PR, but I don't have permission. I'm wondering if someone else can make this change.

```
---
 mempack.go | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/mempack.go b/mempack.go
index 337e67d..26f96be 100644
--- a/mempack.go
+++ b/mempack.go
@@ -6,7 +6,7 @@ package git

 extern int git_mempack_new(git_odb_backend **out);
 extern int git_mempack_dump(git_buf *pack, git_repository *repo, git_odb_backend *backend);
-extern void git_mempack_reset(git_odb_backend *backend);
+extern int git_mempack_reset(git_odb_backend *backend);
 extern void _go_git_odb_backend_free(git_odb_backend *backend);
 */
 import "C"
--
```

### Reproduction steps
`go test`

### Expected behavior
Tests pass.

### Actual behavior
```
Step 18/19 : RUN go test
 ---> Running in 8570ffd11a46
# github.com/libgit2/git2go
./mempack.go:9:13: error: conflicting types for 'git_mempack_reset'
 extern void git_mempack_reset(git_odb_backend *backend);
             ^~~~~~~~~~~~~~~~~
In file included from ./mempack.go:5:0:
/usr/include/git2/sys/mempack.h:83:17: note: previous declaration of 'git_mempack_reset' was here
 GIT_EXTERN(int) git_mempack_reset(git_odb_backend *backend);
                 ^~~~~~~~~~~~~~~~~
FAIL	github.com/libgit2/git2go [build failed]
The command '/bin/sh -c go test' returned a non-zero code: 2
```

### Version of libgit2 (release number or SHA1)
https://github.com/libgit2/libgit2/blob/559751714b23f63ead726e9a773be8b07e3551ae/src/odb_mempack.c#L128

### Operating system(s) tested
`ubuntu:18.04`

PR #534, opened 2020-02-13
Uprev libgit2

lhchavez:
This uprevs libgit2 to the latest and greatest. Notably,

* Fixes the interface of `git_mempack_reset`, since it now returns an
  `int` instead of being `void`.

Fixes: #533


lhchavez:
@lydza can you take a look?

Issue #553, opened 2020-03-23
Cred type finalization can cause `SIGSEGV`


mstergianis:
Hi there,

In using the library I was able to cause a panic in the go runtime. It appears as though a Cred is being freed twice. It appears as though when you use a cred for ssh, libgit2 cleans up the cred that it creates, for example see https://github.com/libgit2/libgit2/blob/172239021f7ba04fe7327647b213799853a9eb89/src/transports/ssh.c#L684-L685. However, git.Cred structs have a finalizer attached to them by default https://github.com/libgit2/git2go/blob/a32375a86063768507a01b32cc3d868f4c2ffa9c/credentials.go#L28-L32.

```golang
package main

import (
	"errors"
	"io/ioutil"
	"log"
	"runtime"

	git "github.com/libgit2/git2go/v29"
)

func main() {
	cloneOptions := &git.CloneOptions{
		FetchOptions: &git.FetchOptions{
			RemoteCallbacks: git.RemoteCallbacks{
				CredentialsCallback:      credCallbackGenerator(false), // change the bool to true to prevent panic
				CertificateCheckCallback: certificateCheckCallback,
			},
		},
	}

	tempDir, err := ioutil.TempDir("./", "cloned-repo-")
	must(err)

	repo, err := git.Clone("git@github.com:githubtraining/hellogitworld.git", tempDir, cloneOptions)
	must(err)

	runtime.GC()
	log.Println(repo)
}

func must(err error) {
	if err != nil {
		panic(err)
	}
}

func credCallbackGenerator(shouldUnsetFinalize bool) func(url string, username string, allowedTypes git.CredType) (*git.Cred, error) {
	return func(url string, username string, allowedTypes git.CredType) (cred *git.Cred, err error) {
		if (allowedTypes & git.CredTypeSshKey) == git.CredTypeSshKey {
			cred, err = git.NewCredSshKeyFromAgent(username)
		} else {
			cred, err = nil, errors.New("only unsupported allowed types")
		}

		if shouldUnsetFinalize && cred != nil {
			runtime.SetFinalizer(cred, nil)
		}

		return cred, err
	}
}

func certificateCheckCallback(cert *git.Certificate, valid bool, hostname string) git.ErrorCode {
	return 0
}
```

On my machine, this exhibits the behaviour I am describing.

lhchavez:
[...]

PR #555, opened 2020-03-26
fix SIGSEGV on double free for Cred object

vladimir-buzuev:
fix for https://github.com/libgit2/git2go/issues/553

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -254,6 +254,8 @@ func credentialsCallback(_cred **C.git_cred, _url *C.char, _username_from_url *C
 		return C.GIT_EUSER
 	}
 	if cred != nil {
+		// we are transferring ownership to libgit, have to disable  
+		runtime.SetFinalizer(cred, nil)
 		*_cred = cred.ptr
 	}
 	return 0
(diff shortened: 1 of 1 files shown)


remote.go
```diff
@@ -254,6 +254,8 @@ func credentialsCallback(_cred **C.git_cred, _url *C.char, _username_from_url *C
 		return C.GIT_EUSER
 	}
 	if cred != nil {
+		// we are transferring ownership to libgit, have to disable  
+		runtime.SetFinalizer(cred, nil)
```
lhchavez: can this also do

```go
cred.ptr = nil
```

afterwards? this should prevent a (theoretical) use-after-free and will change it instead to an immediate crash which should aid in debuggability if the situation were to occur.

vladimir-buzuev: OK, I can add that. But, `cred` is a local obj/var, it's not going to be used from any other places...

lhchavez: theoretically the caller could still be holding a reference to the returned `cred` object from the callback. it's a bit paranoid, but it's still a possibility.

PR #577, opened 2020-05-08
Add a debug allocator

lhchavez:
This change allows debugging all the allocations performed by libgit2.
To enable, run the `./script/leak_detector.py` script (as root), which
makes a UNIX seqpacket socket listen at `/run/git2go_alloc.sock`. Once
that is running, set the environment variable GIT2GO_DEBUG_ALLOCATOR=1
and run the git2go program. Once the program exits, a leak summary will
be printed.

In order to reduce the amount of noise due to the static allocations
performed by libgit2, it is recommended to call `git.Shutdown()` before
exiting the program.

### DIFF
--- a/debug_allocator.c
+++ b/debug_allocator.c
@@ -1,0 +1,193 @@
+#include "_cgo_export.h"
+
+#include <assert.h>
+#include <execinfo.h>
+#include <stdio.h>
+#include <stdint.h>
+#include <stdlib.h>
+#include <sys/socket.h>
+#include <sys/uio.h>
+#include <sys/un.h>
+
+#include <git2.h>
+#include <git2/common.h>
+#include <git2/sys/alloc.h>
+
+static git_allocator _go_git_system_allocator;
+static git_allocator _go_git_debug_allocator;
+
+static int __alloc_fd = -1;
+typedef struct __attribute__ ((__packed__)) {
+	uint32_t type;
+	int32_t line;
+	uint64_t ptr;
+	uint64_t len;
+	uint64_t filelen;
+	uint64_t btlen;
+} __alloc_message;
+static_assert(sizeof(__alloc_message) == 40, "Unexpected size of __alloc_message");
+
+static void log_alloc_event(char type, const void* ptr, size_t len, const char *file, int line) {
+	void *btaddr[16];
+	__alloc_message msg = {
+		.type = type,
+		.line = line,
+		.ptr = (uint64_t)ptr,
+		.len = len,
+		.filelen = (file ? strlen(file) : 0),
+		.btlen = 0,
+	};
+	struct iovec iov[18] = {
+		{.iov_base = &msg, .iov_len = sizeof(msg)},
+		{.iov_base = (char *)file, .iov_len = msg.filelen},
+	};
+	ssize_t expected = iov[0].iov_len + iov[1].iov_len;
+	char **strings = NULL;
+	int iov_count = 2;
+
+	if (type == 'A' || type == 'R') {
+		size_t i;
+		msg.btlen = backtrace(btaddr, 16);
+		strings = backtrace_symbols(btaddr, (int)msg.btlen);
+
+		for (i = 0; i < msg.btlen; ++i) {
+			iov[iov_count].iov_base = strings[i];
+			iov[iov_count].iov_len = strlen(strings[i]) + 1;
+			expected += iov[iov_count].iov_len;
+			++iov_count;
+		}
+	}
+	if (writev(__alloc_fd, iov, iov_count) != expected) {
+		perror("writev");
+		abort();
+	}
+	free(strings);
+}
+
+static void *_go_git_debug_allocator__malloc(size_t len, const char *file, int line)
+{
+	void *ptr = _go_git_system_allocator.gmalloc(len, file, line);
+	if (ptr)
+		log_alloc_event('A', ptr, len, file, line);
+	return ptr;
+}
+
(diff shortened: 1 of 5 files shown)


suhaibmujahid:
Thank you. This will be very helpful for me in debugging a memory leak issue that I'm facing.

I tried to run the following on Mac OS:
```bash
python3 ./script/leak_detector.py
```

However, I got the following error:
```py
Traceback (most recent call last):
  File "./script/leak_detector.py", line 132, in <module>
    main()
  File "./script/leak_detector.py", line 122, in main
    with socket.socket(socket.AF_UNIX, socket.SOCK_SEQPACKET) as sock:
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/socket.py", line 151, in __init__
    _socket.socket.__init__(self, family, type, proto, fileno)
OSError: [Errno 43] Protocol not supported
```

lhchavez:
huh, [`socket(2)`](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/socket.2.html) on the OS X manpages claims to support it. oh well, try applying this patch and see if that helps:

```patch
diff --git a/debug_allocator.c b/debug_allocator.c
index 56a16ec..eff9abb 100644
--- a/debug_allocator.c
+++ b/debug_allocator.c
@@ -153,7 +153,7 @@ int _go_git_setup_debug_allocator()
        struct sockaddr_un name = {};
        int error;
 
-       __alloc_fd = socket(AF_UNIX, SOCK_SEQPACKET, 0);
+       __alloc_fd = socket(AF_UNIX, SOCK_DGRAM, 0);
        if (__alloc_fd == -1) {
                perror("socket");
                return -1;
diff --git a/script/leak_detector.py b/script/leak_detector.py
index ec8a278..72c338f 100755
--- a/script/leak_detector.py
+++ b/script/leak_detector.py
@@ -92,14 +92,18 @@ def _handle_connection(conn: socket.socket) -> None:
     """Handle a single connection."""
 
     live_allocations: Dict[int, Allocation] = {}
-    with conn:
-        for message_type, allocation in _receive_allocation_messages(conn):
-            if message_type in ('A', 'R'):
-                live_allocations[allocation.ptr] = allocation
-            elif message_type == 'D':
-                del live_allocations[allocation.ptr]
-            else:
-                raise Exception(f'Unknown message type "{message_type}"')
+    try:
+        print('Capturing allocations, press Ctrl+C to stop...')
+        with conn:
+            for message_type, allocation in _receive_allocation_messages(conn):
+                if message_type in ('A', 'R'):
+                    live_allocations[allocation.ptr] = allocation
+                elif message_type == 'D':
+                    del live_allocations[allocation.ptr]
+                else:
+                    raise Exception(f'Unknown message type "{message_type}"')
+    except KeyboardInterrupt:
+        pass
     _process_leaked_allocations(live_allocations)
 
 
@@ -119,13 +123,10 @@ def main() -> None:
     except FileNotFoundError:
         pass
 
-    with socket.socket(socket.AF_UNIX, socket.SOCK_SEQPACKET) as sock:
+    with socket.socket(socket.AF_UNIX, socket.SOCK_DGRAM) as sock:
         sock.bind(args.socket_path)
         os.chmod(args.socket_path, 0o666)
-        sock.listen(1)
-        while True:
-            conn, _ = sock.accept()
-            _handle_connection(conn)
+        _handle_connection(sock)
 
 
 if __name__ == '__main__':
```

suhaibmujahid:
Thank you @lhchavez for the patch. I applied it and it seems that it fixed the unsupported protocol issue.

Now I'm getting the following error:
```py
Traceback (most recent call last):
  File "./script/leak_detector.py", line 134, in <module>
    main()
  File "./script/leak_detector.py", line 128, in main
    sock.bind(args.socket_path)
FileNotFoundError: [Errno 2] No such file or directory
```

This error because the script cannot create `/run/git2go_alloc.sock` since `/run` does not exist in MacOS and it cannot be created:
```sh
mkdir: /run: Read-only file system
```

The socket path can be changed in the python script through the argument `socket_path`. However, it is hardcoded in [`debug_allocator.c:163`](https://github.com/libgit2/git2go/blob/a41dbde543f13543cd0b7660fb98b5803d8f593c/debug_allocator.c#L163). It will be nice if it can be changed, through an environment variable for example.

lhchavez:
> Thank you @lhchavez for the patch. I applied it and it seems that it fixed the unsupported protocol issue.
> 
> Now I'm getting the following error:
> 
> ```python
> Traceback (most recent call last):
>   File "./script/leak_detector.py", line 134, in <module>
>     main()
>   File "./script/leak_detector.py", line 128, in main
>     sock.bind(args.socket_path)
> FileNotFoundError: [Errno 2] No such file or directory
> ```
> 
> This error because the script cannot create `/run/git2go_alloc.sock` since `/run` does not exist in MacOS and it cannot be created:
> 
> ```shell
> mkdir: /run: Read-only file system
> ```
> 
> The socket path can be changed in the python script through the argument `socket_path`. However, it is hardcoded in [`debug_allocator.c:163`](https://github.com/libgit2/git2go/blob/a41dbde543f13543cd0b7660fb98b5803d8f593c/debug_allocator.c#L163). It will be nice if it can be changed, through an environment variable for example.

I don't know why I was making the code be more clever than it needed. The latest version uses a text file format that can be either printed out as a regular file, or can be slurped by a FIFO (no more overcomplicated sockets) with the Python script.

suhaibmujahid:
Works perfectly. 

**A small note:** If you call  `./script/leak_detector.py` without the flag `--pipe`, it will not create a new file, and it will raise an exception if the log file does not exist.

PR #578, opened 2020-05-08
Add a way to cleanly shut down the library

lhchavez:
This change adds the Shutdown() method, so that the library can be
cleanly shut down. This helps significanly reduce the amount of noise in
the leak detector.


suhaibmujahid:
Thank you.  I used your patch and it helped me.

PR #582, opened 2020-05-09
Rename methods with Get prefix

suhaibmujahid:
It is not Go idiomatic to put Get into the getter's name ([see here](https://golang.org/doc/effective_go.html#Getters)).

Based on the discussion with @lhchavez on Slack. He suggested the renaming and adding backward compatibility.

### DIFF
--- a/diff.go
+++ b/diff.go
@@ -144,7 +144,7 @@ func (diff *Diff) NumDeltas() (int, error) {
 	return ret, nil
 }
 
-func (diff *Diff) GetDelta(index int) (DiffDelta, error) {
+func (diff *Diff) Delta(index int) (DiffDelta, error) {
 	if diff.ptr == nil {
 		return DiffDelta{}, ErrInvalid
 	}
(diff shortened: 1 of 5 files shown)


diff.go
```diff
@@ -154,6 +154,11 @@ func (diff *Diff) GetDelta(index int) (DiffDelta, error) {
 	return ret, nil
 }
 
+// deprecated: You should use `Diff.GetDelta()` instead.
```
lhchavez, line 157:
```suggestion
// deprecated: You should use `Diff.Delta()` instead.
```

index.go
```diff
@@ -553,6 +553,11 @@ func (v *Index) GetConflict(path string) (IndexConflict, error) {
 	return ret, nil
 }
 
+// deprecated: You should use `Index.GetConflict()` instead.
```
lhchavez, line 556:
```suggestion
// deprecated: You should use `Index.Conflict()` instead.
```

suhaibmujahid:
Thank you for the corrections 👍

suhaibmujahid:
Done!

PR #614, opened 2020-06-01
Add support for parsing git trailers

lollipopman:
Adds a wrapper for git_message_trailers which returns a slice of trailer
structs.

First pull request for git2go & first use of cgo, criticism welcomed, thanks!

### DIFF
--- a/message.go
+++ b/message.go
@@ -1,0 +1,40 @@
+package git
+
+/*
+#include <git2.h>
+*/
+import "C"
+import (
+	"runtime"
+	"unsafe"
+)
+
+type Trailer struct {
+	key   string
+	value string
+}
+
+func MessageTrailers(message string) ([]Trailer, error) {
+
+	var trailersC C.git_message_trailer_array
+
+	messageC := C.CString(message)
+	defer C.free(unsafe.Pointer(messageC))
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ecode := C.git_message_trailers(&trailersC, messageC)
+	if ecode < 0 {
+		return nil, MakeGitError(ecode)
+	}
+	defer C.git_message_trailer_array_free(&trailersC)
+	trailers := make([]Trailer, trailersC.count)
+	var trailer *C.git_message_trailer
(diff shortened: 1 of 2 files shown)


message.go
```diff
@@ -0,0 +1,40 @@
...
+	"runtime"
+	"unsafe"
+)
+
+type Trailer struct {
+	key   string
+	value string
+}
+
+func MessageTrailers(message string) ([]Trailer, error) {
```
lhchavez, line 17: Can the docstrings from [libgit2](https://libgit2.org/libgit2/#HEAD/group/message/git_message_trailers) be copied (+adapted) as the golang docstrings? For example, this could be

```go
// MessageTrailers parses trailers out of a message. Trailers are key/value
// pairs in the last paragraph of a message, not including any patches or
// conflicts that may be present.
```

Same for `Trailer` above in L12 (from https://github.com/libgit2/libgit2/blob/849f371ea89257512ef505d0094d77b1ec812c53/include/git2/message.h#L41)

message.go
```diff
@@ -0,0 +1,40 @@
...
+
```
lhchavez, line 18: minor nit:

```suggestion
```

(if this was copied from another file, let me know which one and we'll fix that later).

message.go
```diff
@@ -0,0 +1,40 @@
...
+	defer runtime.UnlockOSThread()
+
+	ecode := C.git_message_trailers(&trailersC, messageC)
+	if ecode < 0 {
+		return nil, MakeGitError(ecode)
+	}
+	defer C.git_message_trailer_array_free(&trailersC)
+	trailers := make([]Trailer, trailersC.count)
+	var trailer *C.git_message_trailer
+	for i, p := 0, uintptr(unsafe.Pointer(trailersC.trailers)); i < int(trailersC.count); p += unsafe.Sizeof(C.git_message_trailer{}) {
```
lhchavez, line 34: maybe

```suggestion
	for i, p := 0, uintptr(unsafe.Pointer(trailersC.trailers)); i < int(trailersC.count); i, p = i + 1, p + unsafe.Sizeof(C.git_message_trailer{}) {
```

for a tiny bit more symmetry between the initialization and post-steps.

lhchavez:
> @lhchavez thanks for your detailed review, I believe I addressed all of you feedback, but please let me know if there is anything I missed!

it's all good! thanks for doing this!

PR #615, opened 2020-06-04
Provide missing merge flags

takuji:
This change adds two missing merge flags MergeTreeSkipREUC and MergeTreeNoRecursive.
Thanks.


merge.go
```diff
@@ -126,6 +126,13 @@ const (
 	// continue resolving conflicts.  The merge operation will fail with
 	// GIT_EMERGECONFLICT and no index will be returned.
 	MergeTreeFailOnConflict MergeTreeFlag = C.GIT_MERGE_FAIL_ON_CONFLICT
+	// Do not write the REUC extension on the generated index
```
lhchavez, line 129: i'll fix the other docstrings on another change, but would it be possible for the added docstrings to follow Go's convention of starting with the name they are describing?

```go
// MergeTreeSkipREUC does not write ...
```

PR #621, opened 2020-06-21
Revamp the ways in which the library can be built

lhchavez:
This change allows to link the system version of libgit2 statically.
Since `-tags static` is already used for the bundled version of the
library and to avoid breaking old workflows, `-tags
static,system_libgit2` is now used to select that.

This means that the valid combinations are:

| Flag                          | Effect                                        |
|-------------------------------|-----------------------------------------------|
| _No flags_                    | Dynamically-linked against the system libgit2 |
| `-tags static,system_libgit2` | Statically-linked against the system libgit2  |
| `-tags static`                | Statically-linked against the bundled libgit2 |

Note that there is no way to express dynamically linking against the
bundled libgit2 because that makes very little sense, since the binaries
wouldn't be able to be distributed. If that's still desired, the
`PKG_CONFIG_PATH` environment variable can set before building the code.
[`Makefile`](https://github.com/libgit2/git2go/blob/master/Makefile) has
an example of how it is used in the CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,9 +49,17 @@ When linking dynamically against a released version of libgit2, install it via y
 import "github.com/libgit2/git2go/v30"
 ```
 
-### Master branch, or static linking
+### Versioned branch, static linking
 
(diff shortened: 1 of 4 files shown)


suhaibmujahid:
Thank you for this PR. This will eliminate the workaround that I do to build statically because I'm using the go module mod in my project.

lhchavez:
awesome, seems like there is some demand for this. I changed the `system` tag since it's too vague  (and collision-prone) in favor of `system_libgit2`. Let's land this!

PR #623, opened 2020-07-10
FetchOptions: add ability to specify ProxyOptions

lollipopman:
Prior to this change you could not specifiy proxy options on the
FetchOptions struct, which made it impossible to specify a proxy for an
initial clone. This change adds the ProxyOptions to the FetchOptions
struct so you can go through a proxy when cloning.

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -697,7 +697,6 @@ func populateFetchOptions(options *C.git_fetch_options, opts *FetchOptions) {
 	options.custom_headers = C.git_strarray{}
 	options.custom_headers.count = C.size_t(len(opts.Headers))
 	options.custom_headers.strings = makeCStringsFromStrings(opts.Headers)
-	options.proxy_opts = C.git_proxy_options{}
 	populateProxyOptions(&options.proxy_opts, &opts.ProxyOptions)
 }
 
--- a/remote.go
+++ b/remote.go
@@ -117,6 +117,9 @@ type FetchOptions struct {
 
 	// Headers are extra headers for the fetch operation.
 	Headers []string
+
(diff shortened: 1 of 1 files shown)


remote.go
```diff
@@ -117,6 +117,9 @@ type FetchOptions struct {
 
 	// Headers are extra headers for the fetch operation.
 	Headers []string
+
+	// Proxy options to use for this fetch operation
+	*ProxyOptions
```
lhchavez, line 122: it may be better to avoid [embedding](https://golang.org/doc/effective_go.html#embedding) `ProxyOptions`, and also use a value instead of a pointer type since the zero value for `ProxyOptions` is well-defined and matches the semantics of not specifying it (also makes it consistent with all the other fields of `FetchOptions`):

```suggestion
	// ProxyOptions are the proxy options to use for this fetch operation.
	ProxyOptions ProxyOptions
```

remote.go
```diff
@@ -694,6 +697,8 @@ func populateFetchOptions(options *C.git_fetch_options, opts *FetchOptions) {
 	options.custom_headers = C.git_strarray{}
 	options.custom_headers.count = C.size_t(len(opts.Headers))
 	options.custom_headers.strings = makeCStringsFromStrings(opts.Headers)
+	options.proxy_opts = C.git_proxy_options{}
```
lhchavez, line 700: both `git_fetch_init_options()` and `populateProxyOptions()` take care of this, so it can be safely omitted.

lollipopman:
Thanks for the detailed feedback @lhchavez, I believe I have addressed both of your good suggestions, let me know if you what to see any additional changes, thanks!

PR #624, opened 2020-07-23
Fix installation of libgit2 into wrong libdir

pks-t:
Upstream libgit2 has migrated to use the GNUInstallDirs module to decide
where things will be installed to by default. While it improves
consistency with the host system, we're not really after that when
building the vendored libgit2 library, and in fact libgit2.{a,pc} may
now be installed into a different directory than before as it started to
depend on the host platform.

Fix this by explicitly specifying that we want the library and pkgconfig
file to be installed into a plain "lib/" directory.


pks-t:
We may also want to backport this to the v30 branch as it currently breaks install-static on some (most?) platforms completely.

Issue #426, opened 2018-01-26
Add support for git_blob_is_binary


krylovsk:
Add support for [git_blob_is_binary](https://libgit2.github.com/libgit2/#HEAD/group/blob/git_blob_is_binary)

PR #625, opened 2020-07-29
Add support for git_blob_is_binary

vvatanabe:
This adds IsBinary() func to Blob struct, which simply returns the result of git_blob_is_binary function.
Refs: https://libgit2.org/libgit2/#HEAD/group/blob/git_blob_is_binary
Issue: Add support for git_blob_is_binary #426

Fixes: #426

### DIFF
--- a/blob.go
+++ b/blob.go
@@ -40,6 +40,12 @@ func (v *Blob) Contents() []byte {
 	return goBytes
 }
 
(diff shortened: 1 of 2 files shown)


vvatanabe:
I have extended a test in blob_test.go. It works.

PR #626, opened 2020-08-12
Add support for creating signed commits and signing commits during a rebase

mbfr:
This all works but I have a few questions about submitting it:

- I'm targeting this at maser but due to package versions available in other distros I'm guessing I will still have to maintain a fork if I want this to be able to use with v29 and below (eg, pre-1.0 which a lot of distros don't have)?
- There's no dependencies in the go.mod file, but I obviously had to add a dependency on the x/crypto module to add unit tests, note sure if that's going to be a problem?
- In libgit2 `git_commit_create_with_signature` is in the 'repository' block of functions in the documentation, but practically it's very difficult to use like that (you would need to look up the commit you were using to get the commit content anyway) so I added it to the `Commit` struct instead. The other sensible way would possibly be to add it as a method on the `Repository` struct, make it take a commit ID, and make it looks up the commit to read the content, but then it's just a very roundabout way of doing it
- The `RebaseOptions.fromC` does not load the function pointer for `git_commit_signing_cb` (there is a note about it), but I don't know if that use case would ever come up. If it was, there would need to be another bridge function in C that called that function pointer.
- I wasn't really sure on how to actually use the commit that you get from `git_commit_create_with_signature` and I can't find any examples, so what I did in the tests (and what I would probably do if I was using this in the real world) is to create the signed commit, then just re-point head at that signed commit.
- I added a utility method `WithSignatureUsing` to the `Commit` to allow passing a callback to create the signature for the commit, this isn't a 1:1 mapping with the libgit2 api obviously but I figure it's what 99% of people using this library will write themselves anyway


mbfr:
It looks like the tests are failing because of the module dependencies, which I suppose is why there were no dependencies there already. 1.9 hasn't had any releases for over 2 years and 1.10 hasn't had a release for a year and a half, is there any reason they're still in the CI? If they do have to be supported then I'm open to suggestions on how to write tests for this functionality without adding any dependencies

rebase.go
```diff
@@ -69,6 +71,59 @@ func newRebaseOperationFromC(c *C.git_rebase_operation) *RebaseOperation {
...
+	}
+
+	fillBuf := func(bufData string, buf *C.git_buf) error {
+		clen := C.size_t(len(bufData))
+		cstr := unsafe.Pointer(C.CString(bufData))
+
+		// over-assign by a byte (see below)
+		if int(C.git_buf_grow(buf, clen+1)) != 0 {
+			return errors.New("could not grow buffer")
+		}
```
lhchavez, line 99: `git_buf_grow()` internally calls `git__realloc`, which calls [`realloc(3)`](https://linux.die.net/man/3/realloc), which says that "If the new size is larger than the old size, the added memory will _not_ be initialized.". This means that it's not guaranteed that the very last byte in the buffer is going to be a null byte after L107.

Would it make sense to call [`git_buf_sanitize`](https://github.com/libgit2/libgit2/blob/c71321a099373753c22055921c8f538afed5ebb6/src/buffer.c#L143-L150) to ensure the null byte is there?

rebase.go
```diff
@@ -69,6 +71,59 @@ func newRebaseOperationFromC(c *C.git_rebase_operation) *RebaseOperation {
...
+
+		if int(C.git_buf_set(buf, cstr, clen)) != 0 {
+			return errors.New("could not set buffer")
+		}
+
+		// git_buf_set sets 'size' to the 'size' of the buffer to 'clen', but we want it to be clen+1 because after returning it asserts that the buffer ends with a null byte, which Go strings don't
```
lhchavez, line 105: huh, [`C.CString()`] says it does return a NULL-terminated C-compatible string: https://blog.golang.org/cgo#TOC_2.

rebase.go
```diff
@@ -101,21 +157,29 @@ func rebaseOptionsFromC(opts *C.git_rebase_options) RebaseOptions {
 		RewriteNotesRef: C.GoString(opts.rewrite_notes_ref),
 		MergeOptions:    mergeOptionsFromC(&opts.merge_options),
 		CheckoutOptions: checkoutOptionsFromC(&opts.checkout_options),
+		// 	TODO: is it possible to take the callback from C and have it be meaningful in Go? Would we ever want to do it?
```
lhchavez, line 160: this can be removed!

rebase.go
```diff
@@ -101,21 +157,29 @@ func rebaseOptionsFromC(opts *C.git_rebase_options) RebaseOptions {
...
 		quiet:             C.int(ro.Quiet),
 		inmemory:          C.int(ro.InMemory),
 		rewrite_notes_ref: mapEmptyStringToNull(ro.RewriteNotesRef),
 		merge_options:     *ro.MergeOptions.toC(),
 		checkout_options:  *ro.CheckoutOptions.toC(),
 	}
+
+	if ro.SigningCallback != nil {
+		C._go_git_populate_commit_sign_cb(opts)
+		opts.payload = pointerHandles.Track(*ro)
```
lhchavez, line 179:
```suggestion
		opts.payload = pointerHandles.Track(ro)
```

rebase_test.go
```diff
@@ -315,14 +444,40 @@ func commitSomething(repo *Repository, something, content string) (*Oid, error)
 	}
 	defer newTree.Free()
 
-	if err != nil {
-		return nil, err
-	}
```
lhchavez, line 320: can this block be kept?

rebase_test.go
```diff
@@ -130,19 +135,139 @@ func TestRebaseNoConflicts(t *testing.T) {
...
+	if err != nil {
+		t.Logf("Commit is not signed correctly\n%s", commit.RawHeader()+"\n"+commit.RawMessage())
+		return err
+	}
+
+	return nil
 }
 
 // Utils
 func setupRepoForRebase(repo *Repository, masterCommit, branchName string) error {
```
lhchavez, line 258: it's probably okay to always ask for the `commitOpts` to avoid having two flavors of the utility functions. Tests don't need to have backwards-compatibility guarantees.

rebase_test.go
```diff
@@ -130,19 +135,139 @@ func TestRebaseNoConflicts(t *testing.T) {
...
+	head, err := headCommit(repo)
+	checkFatal(t, err)
+
+	defer head.Free()
+
+	parent := head.Parent(0)
+	defer parent.Free()
+
+	err = checkCommitSigned(t, entity, parent)
+	checkFatal(t, err)
```
lhchavez, line 230: IIUC this can be simplified to:

```suggestion
	parent := head
```

rebase_test.go
```diff
@@ -130,19 +135,139 @@ func TestRebaseNoConflicts(t *testing.T) {
...
+
+		err = checkCommitSigned(t, entity, parent)
+		checkFatal(t, err)
+	}
+}
+
+func checkCommitSigned(t *testing.T, entity *openpgp.Entity, commit *Commit) error {
+	signature, signedData, err := commit.ExtractSignature()
+	if err != nil {
+		t.Logf("No signature on commit\n%s", commit.RawHeader()+"\n"+commit.RawMessage())
```
lhchavez, line 244:
```suggestion
		t.Logf("No signature on commit\n%s", commit.ContentToSign())
```

might as well use the new function! Same in L250.

rebase_test.go
```diff
@@ -130,19 +135,139 @@ func TestRebaseNoConflicts(t *testing.T) {
...
+	for parent.ParentCount() != 0 {
+		parent = parent.Parent(0)
+		defer parent.Free()
+
+		err = checkCommitSigned(t, entity, parent)
+		checkFatal(t, err)
+	}
+}
+
+func checkCommitSigned(t *testing.T, entity *openpgp.Entity, commit *Commit) error {
```
lhchavez, line 241: maybe add a

```go
t.Helper()
```

as the first line of this function: https://golang.org/pkg/testing/#T.Helper

rebase_test.go
```diff
@@ -130,19 +135,139 @@ func TestRebaseNoConflicts(t *testing.T) {
...
+	defer parent.Free()
+
+	err = checkCommitSigned(t, entity, parent)
+	checkFatal(t, err)
+
+	for parent.ParentCount() != 0 {
+		parent = parent.Parent(0)
+		defer parent.Free()
+
+		err = checkCommitSigned(t, entity, parent)
```
lhchavez, line 236: given that this is only going to be used once, might as well inline this function

commit.go
```diff
@@ -40,6 +40,62 @@ func (c *Commit) RawMessage() string {
...
+
+	var csf *C.char = nil
+	if signatureField != "" {
+		csf = C.CString(signatureField)
+	}
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ret := C.git_commit_create_with_signature(
```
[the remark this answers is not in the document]
mbfr: There isn't really a lot of consistency with other functions, sometimes it's called `code`, sometimes `ret`, sometimes, `ecode`, sometimes `cerr`, etc.

commit.go
```diff
@@ -40,6 +40,62 @@ func (c *Commit) RawMessage() string {
...
+
+// WithSignature creates a new signed commit from the given signature and signature field
+func (c *Commit) WithSignature(signature string, signatureField string) (*Oid, error) {
+	totalCommit := c.ContentToSign()
+
+	oid := new(Oid)
+
+	var csf *C.char = nil
+	if signatureField != "" {
+		csf = C.CString(signatureField)
```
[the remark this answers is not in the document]
mbfr: Added

commit.go
```diff
@@ -40,6 +40,62 @@ func (c *Commit) RawMessage() string {
...
+	}
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ret := C.git_commit_create_with_signature(
+		oid.toC(),
+		c.Owner().ptr,
+		C.CString(totalCommit),
+		C.CString(signature),
```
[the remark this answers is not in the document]
mbfr: Added

rebase.go
```diff
@@ -69,6 +71,59 @@ func newRebaseOperationFromC(c *C.git_rebase_operation) *RebaseOperation {
...
+
+	if opts.SigningCallback == nil {
+		return C.GIT_PASSTHROUGH
+	}
+
+	commitContent := C.GoString(_commit_content)
+
+	signature, signatureField, err := opts.SigningCallback(commitContent)
+	if err != nil {
+		return C.int(-1)
```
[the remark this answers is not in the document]
mbfr: Good idea, added

rebase.go
```diff
@@ -69,6 +71,59 @@ func newRebaseOperationFromC(c *C.git_rebase_operation) *RebaseOperation {
...
+	}
+
+	fillBuf := func(bufData string, buf *C.git_buf) error {
+		clen := C.size_t(len(bufData))
+		cstr := unsafe.Pointer(C.CString(bufData))
```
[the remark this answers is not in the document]
mbfr: Fixed

rebase.go
mbfr: I did have this doing memset() to null bytes but figured that it was probably done by libgit internally, apparently not...I think I'll just make it do that again to make sure that it has a null byte and to avoid any issues with sizes being incorrect

rebase.go
mbfr: I suppose the easier thing to do would be to just do clen+1 in the git_buf_set as well then, fixed

rebase.go
mbfr: Removed

rebase.go
mbfr: Oops, will accept the suggestion

rebase_test.go
mbfr: It's already checked directly above, I don't think it needs to do it again?

rebase_test.go
mbfr: Done

rebase_test.go
mbfr: Added, should make logs a bit nicer

rebase_test.go
mbfr: It's used in 2 places, unless you mean you want to to refactor it so both of those invocations are inside the for loop?

rebase_test.go
mbfr: Ah yes, it misses the case in which the head commit is not signed, fixed

rebase_test.go
mbfr: Fixed this and the other similar ones

rebase.go
```diff
@@ -69,6 +72,61 @@ func newRebaseOperationFromC(c *C.git_rebase_operation) *RebaseOperation {
...
+		// over-assign by a byte (see below)
+		if int(C.git_buf_grow(buf, clen+1)) != 0 {
+			return errors.New("could not grow buffer")
+		}
+
+		C._go_git_buf_fill_null(buf)
+
+		if int(C.git_buf_set(buf, cstr, clen+1)) != 0 {
+			return errors.New("could not set buffer")
+		}
```
lhchavez, line 110: OK, second try:

```suggestion
		C.git_buf_clear(buf)
		if int(C.git_buf_put(buf, cstr, clen)) != 0 {
			return errors.New("could not set buffer")
		}
```

[`git_buf_clear()`](https://github.com/libgit2/libgit2/blob/c71321a099373753c22055921c8f538afed5ebb6/src/buffer.c#L152-L163) will remove whatever is in the buffer, and [`git_buf_put()`](https://github.com/libgit2/libgit2/blob/c71321a099373753c22055921c8f538afed5ebb6/src/buffer.c#L223-L238) grows the buffer AND adds the NULL byte at the end.

wrapper.c
```diff
@@ -6,6 +6,16 @@
 
 typedef int (*gogit_submodule_cbk)(git_submodule *sm, const char *name, void *payload);
 
+void _go_git_buf_fill_null(git_buf *buf)
+{
+  memset(buf->ptr, '\0', buf->asize*sizeof(char));
+}
+
```
lhchavez, line 13:
```suggestion
```

this should be unneeded now.

commit.go
```diff
@@ -40,6 +40,68 @@ func (c *Commit) RawMessage() string {
...
+	return ret
+}
+
+// ContentToSign returns the content that will be passed to a signing function for this commit
+func (c *Commit) ContentToSign() string {
+	return c.RawHeader() + "\n" + c.RawMessage()
+}
+
+// CommitSigningCb defines a function type that takes some data to sign and returns (signature, signature_field, error)
+type CommitSigningCb func(string) (string, string, error)
```
lhchavez, line 55:
```suggestion
// CommitSigningCb defines a function type that takes some data to sign and returns (signature, signature_field, error)
type CommitSigningCb func(string) (signature, signatureField string, err error)
```

welp, i missed this. it should make the godoc a bit better.

commit.go
```diff
@@ -40,6 +40,68 @@ func (c *Commit) RawMessage() string {
...
+	var csf *C.char = nil
+	if signatureField != "" {
+		csf = C.CString(signatureField)
+		defer C.free(unsafe.Pointer(csf))
+	}
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	cTotalCommit:=C.CString(totalCommit)
```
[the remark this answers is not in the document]
mbfr: Fixed

rebase.go
mbfr: It seems like those functions aren't defined in the 'buffer.h' header and are also not specified in the [documentation](https://libgit2.org/libgit2/#HEAD/search/git_buf_), I tried manually adding the `extern ...` but got an undefined reference error. Are they intended only for internal use?

commit.go
mbfr: Fixed this and also changed the name of the type/field to be more appropriate (based on feedback from other PR)

rebase.go
lhchavez: argh, yes u_u they are not marked `GIT_EXTERN`.

but why is it not possible to use

```suggestion
		if int(C.git_buf_set(buf, cstr, clen)) != 0 {
			return errors.New("could not set buffer")
		}
```

? [`git_buf_set`](https://github.com/libgit2/libgit2/blob/c71321a099373753c22055921c8f538afed5ebb6/src/buffer.c#L165-L184) resizes the buffer (plus the space for the NULL byte), and adds the trailing NULL byte. That's even independent of whether the original c-string had a NULL pointer or not. (This time around I did test this suggestion locally, so I'm sure this one does definitely work).

wrapper.c
```diff
@@ -6,6 +6,16 @@
...
+void _go_git_populate_commit_sign_cb(git_rebase_options *opts)
+{
+  opts->signing_cb = (git_commit_signing_cb)commitSignCallback;
```
lhchavez, line 22:
```suggestion
	opts->signing_cb = (git_commit_signing_cb)commitSignCallback;
```

for consistency with the other entries in this file.

commit.go
```diff
@@ -40,6 +40,68 @@ func (c *Commit) RawMessage() string {
 	return ret
 }
 
+func (c *Commit) RawHeader() string {
```
lhchavez, line 43: one more thing i forgot, can the public functions have a docstring? most of the contents can be copied from the C docs.

rebase.go
mbfr: I tried this before and although it works in CI (see: github actions on that commit), when running locally I get this error
```
___go_test_github_com_libgit2_git2go_v30: /build/libgit2/src/libgit2-1.0.1/src/rebase.c:1007: rebase_commit__create: Assertion `git_buf_contains_nul(&commit_signature)' failed.
```

wrapper.c
mbfr: Fixed

commit.go
mbfr: Fixed

rebase.go
lhchavez: interesting, that means that we need to have at least one CI configuration where assertions are enabled.

okay, what about

```suggestion
		// libgit2 requires the contents of the buffer to be NULL-terminated.
		// C.CString() guarantees that is the case, so we can safely copy
		// the NULL terminator.
		if int(C.git_buf_set(buf, cstr, clen+1)) != 0 {
			return errors.New("could not set buffer")
		}
```

mbfr: Fixed (and removed the old wrapper)

PR #628, opened 2020-08-14
Fix null pointer dereference in status.ByIndex

mbfr:
`git_status_byindex` can return a null pointer if there is no statuses.

### DIFF
--- a/status.go
+++ b/status.go
@@ -87,6 +87,8 @@ func (statusList *StatusList) ByIndex(index int) (StatusEntry, error) {
 	}
 	ptr := C.git_status_byindex(statusList.ptr, C.size_t(index))
 	if ptr == nil {
+		runtime.LockOSThread()
+		defer runtime.UnlockOSThread()
 		return StatusEntry{}, MakeGitError(C.int(ErrNotFound))
 	}
 	entry := statusEntryFromC(ptr)
--- a/status.go
+++ b/status.go
@@ -86,6 +86,9 @@ func (statusList *StatusList) ByIndex(index int) (StatusEntry, error) {
 		return StatusEntry{}, ErrInvalid
 	}
 	ptr := C.git_status_byindex(statusList.ptr, C.size_t(index))
+	if ptr == nil {
+		return StatusEntry{}, MakeGitError(C.int(ErrNotFound))
+	}
 	entry := statusEntryFromC(ptr)
 	runtime.KeepAlive(statusList)
 
--- a/status.go
+++ b/status.go
@@ -6,6 +6,7 @@ package git
 import "C"
 
 import (
+	"errors"
 	"runtime"
 	"unsafe"
 )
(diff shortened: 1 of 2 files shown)


status.go
```diff
@@ -86,6 +86,11 @@ func (statusList *StatusList) ByIndex(index int) (StatusEntry, error) {
 		return StatusEntry{}, ErrInvalid
 	}
 	ptr := C.git_status_byindex(statusList.ptr, C.size_t(index))
+	if ptr == nil {
+		runtime.LockOSThread()
+		defer runtime.UnlockOSThread()
+		return StatusEntry{}, MakeGitError(C.int(ErrNotFound))
```
lhchavez, line 92: welp, unfortunately there's no consistency with what these functions return :(

`Blame.HunkByIndex`: https://github.com/libgit2/git2go/blob/462ebd83e0ccba9cd93c05ec12dc3d98064e3d5c/blame.go#L102
`Index.EntryByIndex`: https://github.com/libgit2/git2go/blob/462ebd83e0ccba9cd93c05ec12dc3d98064e3d5c/index.go#L420
`Tree.EntryByIndex`: Does nothing and may also need to be fixed (although since it returns a pointer, presumably it could return `nil` when that is fixed).

But in this case, since libgit2 does not set an error itself, using `MakeGitError()` will call [`giterr_last()`](https://github.com/libgit2/git2go/blob/462ebd83e0ccba9cd93c05ec12dc3d98064e3d5c/git.go#L281), which will most likely return `GIT_OK`, which will cause this error to have an empty message.

so I *think* the more correct solution here is to do the same thing as `Index.EntryByIndex`:

```suggestion
		return StatusEntry{}, fmt.Errorf("Index out of Bounds")
```

status_test.go
```diff
@@ -61,3 +61,30 @@ func TestStatusList(t *testing.T) {
...
+func TestStatusNothing(t *testing.T) {
+	t.Parallel()
+	repo := createTestRepo(t)
+	defer cleanupTestRepo(t, repo)
+
+	seedTestRepo(t, repo)
+
+	opts := &StatusOptions{}
+	opts.Show = StatusShowIndexAndWorkdir
+	opts.Flags = StatusOptIncludeUntracked | StatusOptRenamesHeadToIndex | StatusOptSortCaseSensitively
```
lhchavez, line 74: nit:

```suggestion
	opts := &StatusOptions{
		Show:  StatusShowIndexAndWorkdir,
		Flags: StatusOptIncludeUntracked | StatusOptRenamesHeadToIndex | StatusOptSortCaseSensitively,
	}
```

this is a bit [more idiomatic](https://golang.org/doc/effective_go.html#composite_literals).

status_test.go
```diff
@@ -61,3 +61,30 @@ func TestStatusList(t *testing.T) {
...
+
+	statusList, err := repo.StatusList(opts)
+	checkFatal(t, err)
+
+	entryCount, err := statusList.EntryCount()
+	checkFatal(t, err)
+
+	if entryCount != 0 {
+		t.Error("expected no statuses in empty repo")
```
lhchavez, line 83: nit:

```suggestion
		t.Fatal("expected no statuses in empty repo")
```

if this fails, then the preconditions for the rest of the test case are not met, which will make the rest of the assertions fail, causing log spam.

status_test.go
mbfr: This was probably me looking for a `DefaultStatusOptions` and not finding it then not going back and fixing it

PR #629, opened 2020-08-14
More diff functionality

mbfr:
This PR adds
- The ability to apply a Diff object to the repo
- Support for git_apply_hunk_cb and git_apply_delta_cb callbacks in options for applying the diffs
- The ability to import a diff from a raw buffer (for example, one exported by ToBuf) into a Diff object associated with the repo
- Tests for the above

The test files might be a bit messy but there's quite a lot of stuff to cover.

### DIFF
--- a/diff.go
+++ b/diff.go
@@ -916,10 +916,11 @@ func deltaApplyCallback(_delta *C.git_diff_delta, _payload unsafe.Pointer) C.int
 func DefaultApplyOptions() (*ApplyOptions, error) {
 	opts := C.git_apply_options{}
 
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
 	ecode := C.git_apply_options_init(&opts, C.GIT_APPLY_OPTIONS_VERSION)
 	if int(ecode) != 0 {
-		runtime.LockOSThread()
-		defer runtime.UnlockOSThread()
 
 		return nil, MakeGitError(ecode)
 	}
--- a/diff.go
+++ b/diff.go
@@ -878,6 +878,9 @@ func hunkApplyCallback(_hunk *C.git_diff_hunk, _payload unsafe.Pointer) C.int {
 
 	apply, err := opts.ApplyHunkCallback(&hunk)
 	if err != nil {
+		if gitError, ok := err.(*GitError); ok {
+			return C.int(gitError.Code)
+		}
 		return -1
 	} else if apply {
 		return 0
@@ -901,6 +904,9 @@ func deltaApplyCallback(_delta *C.git_diff_delta, _payload unsafe.Pointer) C.int
 
 	apply, err := opts.ApplyDeltaCallback(&delta)
 	if err != nil {
+		if gitError, ok := err.(*GitError); ok {
+			return C.int(gitError.Code)
+		}
 		return -1
 	} else if apply {
 		return 0
--- a/diff.go
+++ b/diff.go
@@ -850,17 +850,17 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
 	return nil
 }
 
-type ApplyHunkCallback func(*DiffHunk) (apply bool, err error)
-type ApplyDeltaCallback func(*DiffDelta) (apply bool, err error)
+type ApplyHunkCb func(*DiffHunk) (apply bool, err error)
+type ApplyDeltaCb func(*DiffDelta) (apply bool, err error)
 
 // ApplyOptions has 2 callbacks that are called for hunks or deltas
 // If these functions return an error, abort the apply process immediately.
-// If the first resutnr value is true, the delta/hunk will be applied. If it is false, the  delta/hunk will not be applied. In either case, the rest of the apply process will continue.
+// If the first return value is true, the delta/hunk will be applied. If it is false, the  delta/hunk will not be applied. In either case, the rest of the apply process will continue.
 type ApplyOptions struct {
-	Version uint
-	ApplyHunkCallback
-	ApplyDeltaCallback
-	Flags uint
+	Version            uint
+	ApplyHunkCallback  ApplyHunkCb
+	ApplyDeltaCallback ApplyDeltaCb
+	Flags              uint
 }
 
 //export hunkApplyCallback
--- a/diff.go
+++ b/diff.go
@@ -969,10 +969,13 @@ func (v *Repository) ApplyDiff(diff *Diff, location GitApplyLocation, opts *Appl
 func DiffFromBuffer(buffer []byte, repo *Repository) (*Diff, error) {
 	var diff *C.git_diff
 
+	cBuffer := C.CBytes(buffer)
+	defer C.free(unsafe.Pointer(cBuffer))
+
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
-	ecode := C.git_diff_from_buffer(&diff, C.CString(string(buffer)), C.size_t(len(buffer)))
+	ecode := C.git_diff_from_buffer(&diff, (*C.char)(cBuffer), C.size_t(len(buffer)))
 	if ecode < 0 {
 		return nil, MakeGitError(ecode)
 	}
--- a/diff.go
+++ b/diff.go
@@ -865,7 +865,7 @@ type ApplyOptions struct {
 
 //export hunkApplyCallback
 func hunkApplyCallback(_hunk *C.git_diff_hunk, _payload unsafe.Pointer) C.int {
-	opts, ok := pointerHandles.Get(_payload).(ApplyOptions)
+	opts, ok := pointerHandles.Get(_payload).(*ApplyOptions)
 	if !ok {
 		panic("invalid apply options payload")
 	}
@@ -888,7 +888,7 @@ func hunkApplyCallback(_hunk *C.git_diff_hunk, _payload unsafe.Pointer) C.int {
 
 //export deltaApplyCallback
 func deltaApplyCallback(_delta *C.git_diff_delta, _payload unsafe.Pointer) C.int {
-	opts, ok := pointerHandles.Get(_payload).(ApplyOptions)
+	opts, ok := pointerHandles.Get(_payload).(*ApplyOptions)
 	if !ok {
 		panic("invalid apply options payload")
 	}
@@ -929,7 +929,7 @@ func (a *ApplyOptions) toC() *C.git_apply_options {
 
 	if a.ApplyDeltaCallback != nil || a.ApplyHunkCallback != nil {
 		C._go_git_populate_apply_cb(opts)
-		opts.payload = pointerHandles.Track(*a)
+		opts.payload = pointerHandles.Track(a)
 	}
 
 	return opts
--- a/diff.go
+++ b/diff.go
@@ -912,9 +912,6 @@ func deltaApplyCallback(_delta *C.git_diff_delta, _payload unsafe.Pointer) C.int
 func DefaultApplyOptions() (*ApplyOptions, error) {
 	opts := C.git_apply_options{}
 
-	runtime.LockOSThread()
-	defer runtime.UnlockOSThread()
-
 	C._go_git_apply_init_options(&opts)
 
 	return applyOptionsFromC(&opts), nil
--- a/diff.go
+++ b/diff.go
@@ -957,8 +957,11 @@ func (v *Repository) ApplyDiff(diff *Diff, location GitApplyLocation, opts *Appl
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
-	ecode := C.git_apply(v.ptr, diff.ptr, C.git_apply_location_t(location), opts.toC())
+	cOpts := opts.toC()
+	ecode := C.git_apply(v.ptr, diff.ptr, C.git_apply_location_t(location), cOpts)
 	runtime.KeepAlive(v)
+	runtime.KeepAlive(diff)
+	runtime.KeepAlive(cOpts)
 	if ecode < 0 {
 		return MakeGitError(ecode)
 	}
--- a/diff.go
+++ b/diff.go
@@ -849,7 +849,10 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
 	return nil
 }
 
+// ApplyHunkCallback is a callback that will be made per delta (file) when applying a patch.
 type ApplyHunkCallback func(*DiffHunk) (apply bool, err error)
+
+// ApplyDeltaCallback is a callback that will be made per hunk when applying a patch.
 type ApplyDeltaCallback func(*DiffDelta) (apply bool, err error)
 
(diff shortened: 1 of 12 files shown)


diff.go
```diff
@@ -847,3 +849,134 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
...
+	GitApplyLocationIndex   GitApplyLocation = C.GIT_APPLY_LOCATION_INDEX
+	GitApplyLocationBoth    GitApplyLocation = C.GIT_APPLY_LOCATION_BOTH
+)
+
+func (v *Repository) ApplyDiff(diff *Diff, location GitApplyLocation, opts *ApplyOptions) error {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ecode := C.git_apply(v.ptr, diff.ptr, C.git_apply_location_t(location), opts.toC())
+	runtime.KeepAlive(v)
```
lhchavez, line 961:
```suggestion
	runtime.KeepAlive(v)
	runtime.KeepAlive(diff)
	runtime.KeepAlive(cOpts)
```

diff.go
```diff
@@ -847,3 +849,134 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
...
+	return nil
+}
+
+func DiffFromBuffer(buffer []byte, repo *Repository) (*Diff, error) {
+	var diff *C.git_diff
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ecode := C.git_diff_from_buffer(&diff, C.CString(string(buffer)), C.size_t(len(buffer)))
```
lhchavez, line 975: the `C.CString()` needs to be `C.free()`d. Although it might be better to use [`C.CBytes()`](https://golang.org/cmd/cgo/) since the buffer _should_ be able to contain NULL bytes.

diff.go
```diff
@@ -847,3 +849,134 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
...
+type ApplyHunkCallback func(*DiffHunk) (apply bool, err error)
+type ApplyDeltaCallback func(*DiffDelta) (apply bool, err error)
+
+// ApplyOptions has 2 callbacks that are called for hunks or deltas
+// If these functions return an error, abort the apply process immediately.
+// If the first resutnr value is true, the delta/hunk will be applied. If it is false, the  delta/hunk will not be applied. In either case, the rest of the apply process will continue.
+type ApplyOptions struct {
+	Version uint
+	ApplyHunkCallback
+	ApplyDeltaCallback
```
lhchavez, line 862: Is it possible to avoid [embedding](https://golang.org/doc/effective_go.html#embedding) the callbacks into this struct? otherwise it gives the appearance of there being an 'is-a' relationship between `ApplyOptions` and `ApplyHunkCallback`.

diff.go
```diff
@@ -847,3 +849,134 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
...
+	Flags uint
+}
+
+//export hunkApplyCallback
+func hunkApplyCallback(_hunk *C.git_diff_hunk, _payload unsafe.Pointer) C.int {
+	opts, ok := pointerHandles.Get(_payload).(ApplyOptions)
```
lhchavez, line 868: can the `pointerHandles` only store the pointer to the `ApplyOptions`?

```suggestion
	opts, ok := pointerHandles.Get(_payload).(*ApplyOptions)
```

diff.go
```diff
@@ -847,3 +849,134 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
...
+		return 1
+	}
+}
+
+func DefaultApplyOptions() (*ApplyOptions, error) {
+	opts := C.git_apply_options{}
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
```
lhchavez, line 917: since there is no interaction with the thread-local storage in libgit2 (e.g. `git_error`-related stuff), this is not required here.

diff.go
```diff
@@ -847,3 +849,134 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
...
+	C._go_git_apply_init_options(&opts)
```
lhchavez, line 918: other places (like https://github.com/libgit2/git2go/blob/fc6eaf36388841b16ff004e1d48e887d3f9613dc/diff.go#L514) use the function initialization rather than the initialization constant to avoid having to create a wrapper:

```suggestion
	ecode := C.git_apply_options_init(&opts, C.GIT_APPLY_OPTIONS_VERSION)
```

diff.go
```diff
@@ -847,3 +849,134 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
...
+	}
+
+	opts := &C.git_apply_options{
+		version: C.uint(a.Version),
+		flags:   C.uint(a.Flags),
+	}
+
+	if a.ApplyDeltaCallback != nil || a.ApplyHunkCallback != nil {
+		C._go_git_populate_apply_cb(opts)
+		opts.payload = pointerHandles.Track(*a)
```
lhchavez, line 935:
```suggestion
		opts.payload = pointerHandles.Track(a)
```

diff_test.go
```diff
@@ -236,3 +239,307 @@ func TestDiffBlobs(t *testing.T) {
 		t.Fatalf("Bad number of lines iterated")
 	}
 }
+
+func Test_ApplyDiff_Addfile(t *testing.T) {
```
lhchavez, line 243:
```suggestion
func TestApplyDiffAddfile(t *testing.T) {
```

so far, no tests have underscores in their names.

wrapper.c
```diff
@@ -6,6 +6,18 @@
...
+void _go_git_apply_init_options(git_apply_options *options)
+{
+  git_apply_options opts = GIT_APPLY_OPTIONS_INIT;
+  *options = opts;
+}
+
+void _go_git_populate_apply_cb(git_apply_options *options)
+{
+  options->delta_cb = (git_apply_delta_cb)deltaApplyCallback;
+  options->hunk_cb = (git_apply_hunk_cb)hunkApplyCallback;
```
lhchavez, line 18: can this use tabs for consistency?

```suggestion
	options->delta_cb = (git_apply_delta_cb)deltaApplyCallback;
	options->hunk_cb = (git_apply_hunk_cb)hunkApplyCallback;
```

diff.go
```diff
@@ -847,3 +849,134 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
...
+
+	if opts.ApplyHunkCallback == nil {
+		return 0
+	}
+
+	hunk := diffHunkFromC(_hunk)
+
+	apply, err := opts.ApplyHunkCallback(&hunk)
+	if err != nil {
+		return -1
```
lhchavez, line 881: can this also use the same pattern as the other change?

```go
		if gitError, ok := err.(*GitError); ok {
			return C.int(gitError.Code)
		}
		return -1
```

diff.go
```diff
@@ -847,3 +849,134 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
...
+	GitApplyLocationWorkdir GitApplyLocation = C.GIT_APPLY_LOCATION_WORKDIR
+	GitApplyLocationIndex   GitApplyLocation = C.GIT_APPLY_LOCATION_INDEX
+	GitApplyLocationBoth    GitApplyLocation = C.GIT_APPLY_LOCATION_BOTH
+)
+
+func (v *Repository) ApplyDiff(diff *Diff, location GitApplyLocation, opts *ApplyOptions) error {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ecode := C.git_apply(v.ptr, diff.ptr, C.git_apply_location_t(location), opts.toC())
```
lhchavez, line 960: can the `.toC()` call be outlined? all other places use the

```go
cOpts = opts.toC()
...
ecode := C.git_...(cOpts)
```

pattern (except for `Oid`, since those are just cast to unsafe pointers).

diff.go
```diff
@@ -847,3 +849,134 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
...
 }
+
+type ApplyHunkCallback func(*DiffHunk) (apply bool, err error)
+type ApplyDeltaCallback func(*DiffDelta) (apply bool, err error)
+
+// ApplyOptions has 2 callbacks that are called for hunks or deltas
+// If these functions return an error, abort the apply process immediately.
+// If the first resutnr value is true, the delta/hunk will be applied. If it is false, the  delta/hunk will not be applied. In either case, the rest of the apply process will continue.
+type ApplyOptions struct {
+	Version uint
```
lhchavez, line 860:
```suggestion
```

the version is not needed to be exposed. it's only needed internally by libgit2 to ensure that the size of the struct is what it expects.

mbfr: I exposed it because it is exposed in some of the other structs (For example, CherrypickOptions, RebaseOptions, etc). Should I just convert this to a non-exported field instead?

diff.go
mbfr: Fixed

diff.go
mbfr: Fixed

diff.go
mbfr: Fixed

wrapper.c
mbfr: Weird, don't know how that happened

diff.go
mbfr: I didn't realise that, I suppose that's why it passes the size as a separate parameter (so it can do a binary diff?) I'm not sure about the CString, I wasn't able to find out how Go behaves if theres a null byte in the middle of a string so I converted it to a CBytes as recommended

diff.go
mbfr: Fixed

diff.go
mbfr: Fixed

diff.go
mbfr: I thought there was a reason I did it this way but after changing it to call this it works fine :thinking:

diff.go
lhchavez: I'd rather remove it altogether since that version number is just there to account for version drift in C, where there is no way to differentiate between different versions of the same struct. As long as there is a layer that accounts for this (e.g. git2go), it's an implementation detail that's best to hide from the end user.

RevertOptions was recently de-Versionified, so we should probably make the other Options follow suit: https://github.com/libgit2/git2go/commit/4bca045e5aa98b0b791fb467705de0692fe3514f#diff-30188d21ac9afa73021c8dd3ae818448 I'll probably do that at v31 branch creation time to avoid breaking the interface for older folks.

diff.go
```diff
@@ -850,22 +849,22 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
 	return nil
 }
 
-type ApplyHunkCallback func(*DiffHunk) (apply bool, err error)
-type ApplyDeltaCallback func(*DiffDelta) (apply bool, err error)
+type ApplyHunkCb func(*DiffHunk) (apply bool, err error)
+type ApplyDeltaCb func(*DiffDelta) (apply bool, err error)
```
lhchavez, line 853:
```suggestion
type ApplyHunkCallback func(*DiffHunk) (apply bool, err error)
type ApplyDeltaCallback func(*DiffDelta) (apply bool, err error)
```

the older name was better for consistency:

```shell
~/git2go$ git grep 'type.*Cb func' | wc -l
0
~/git2go$ git grep 'type.*Callback func' | wc -l
24
```

in L860-L861, the same name can be used for the field and the type, which is the way to express that something is not to be embedded.

diff.go
```diff
@@ -912,10 +917,13 @@ func deltaApplyCallback(_delta *C.git_diff_delta, _payload unsafe.Pointer) C.int
 func DefaultApplyOptions() (*ApplyOptions, error) {
 	opts := C.git_apply_options{}
 
-	runtime.LockOSThread()
-	defer runtime.UnlockOSThread()
+	ecode := C.git_apply_options_init(&opts, C.GIT_APPLY_OPTIONS_VERSION)
+	if int(ecode) != 0 {
+		runtime.LockOSThread()
+		defer runtime.UnlockOSThread()
```
lhchavez, line 923: Can these be returned to their old place? The reason why they are needed is that when [`runtime.LockOsThread()`](https://golang.org/pkg/runtime/#LockOSThread)/`runtime.UnlockOSThread()` combo is called, it asks Go to guarantee that _only_ this code executes on that thread (since Go is completely free to move stuff around when needed, at any point in time). That's the only way in which it can be guaranteed that if `git_apply_options_init` happens to place any error information in the Thread-local storage, `MakeGitError()` below can still find it.

We may need to tweak `script/check-MakeGitError-thread-lock.go` to _also_ complain if these functions don't happen before any cgo calls in the function to make this less error-prone.

diff.go
```diff
@@ -850,22 +849,22 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
...
-// If the first resutnr value is true, the delta/hunk will be applied. If it is false, the  delta/hunk will not be applied. In either case, the rest of the apply process will continue.
+// If the first return value is true, the delta/hunk will be applied. If it is false, the  delta/hunk will not be applied. In either case, the rest of the apply process will continue.
 type ApplyOptions struct {
-	Version uint
-	ApplyHunkCallback
-	ApplyDeltaCallback
-	Flags uint
+	Version            uint
+	ApplyHunkCallback  ApplyHunkCb
+	ApplyDeltaCallback ApplyDeltaCb
```
lhchavez, line 861:
```suggestion
	ApplyHunkCallback  ApplyHunkCallback
	ApplyDeltaCallback ApplyDeltaCallback
```

(as mentioned above, this is possible)

diff.go
```diff
@@ -847,3 +848,149 @@ func DiffBlobs(oldBlob *Blob, oldAsPath string, newBlob *Blob, newAsPath string,
...
 }
+
+type ApplyHunkCb func(*DiffHunk) (apply bool, err error)
+type ApplyDeltaCb func(*DiffDelta) (apply bool, err error)
+
+// ApplyOptions has 2 callbacks that are called for hunks or deltas
+// If these functions return an error, abort the apply process immediately.
+// If the first return value is true, the delta/hunk will be applied. If it is false, the  delta/hunk will not be applied. In either case, the rest of the apply process will continue.
+type ApplyOptions struct {
+	Version            uint
```
lhchavez, line 859:
```suggestion
```

(from the other thread, in case it gets lost)

diff.go
mbfr: Fixed

diff.go
mbfr: Removed and hardcoded like the other options

diff.go
mbfr: Fixed

PR #642, opened 2020-09-18
merge: Expose recursion limit merge option

pks-t:
The `recursion_limit` merge option provided by libgit2 is currently not
exposed and thus inaccessible to Git2Go users. Let's expose it to let
users control creation of recursive merge bases.


lhchavez (approved):
Thanks!

PR #647, opened 2020-09-28
Add a ReInit function

lollipopman:
libgit2 stores the lookup paths for gitconfig files in its global state.
This means that after a program changes its effective uid libgit2 will
still look for gitconfig files in the home directory of the original
uid. Expose a way to call C.git_libgit2_init so a user can reinitialize
the libgit2 global state.

### DIFF
--- a/git.go
+++ b/git.go
@@ -149,6 +149,17 @@ func Shutdown() {
 	C.git_libgit2_shutdown()
 }
 
+// ReInit reinitializes the global state, this is useful if the effective user
+// id has changed and you want to update the stored search paths for gitconfig
+// files. This function frees any references to objects, so it should be called
+// before any other functions are called.
+func ReInit() {
+	pointerHandles.Clear()
+	C.git_libgit2_shutdown()
+	pointerHandles = NewHandleList()
+	C.git_libgit2_init()
+}
+
 // Oid represents the id for a Git object.
 type Oid [20]byte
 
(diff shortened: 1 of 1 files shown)


git.go
```diff
@@ -149,6 +149,17 @@ func Shutdown() {
...
 
+// ReInit reinitializes the global state, this is useful if the effective user
+// id has changed and you want to update the stored search paths for gitconfig
+// files. This function frees any references to objects, so it should be called
+// before any other functions are called.
+func ReInit() {
+	pointerHandles.Clear()
+	C.git_libgit2_shutdown()
+	pointerHandles = NewHandleList()
+	C.git_libgit2_init()
```
lhchavez, line 160: oh, interesting use case! Although would something like this work?

```suggestion
	Shutdown()
	init()
```

if it doesn't, is there any way in which `init()` can be refactored so that there is less repetition?

lollipopman:
thanks @lhchavez for taking a look, I pushed a commit to reduce the repetition, which seems to work fine, though I am not super happy with the function naming.

lhchavez:
> thanks @lhchavez for taking a look, I pushed a commit to reduce the repetition, which seems to work fine, though I am not super happy with the function naming.

That name is fine. Since it's a private function, if you ever come up with a better alternative, we can change it without fear of breaking anything!

lhchavez (approved):

PR #656, opened 2020-10-09
Use un-deprecated options init functions from libgit2

suhaibmujahid:
This PR move form linking against the deprecated `init_options` functions to the renamed `options_init` functions.

For more context see libgit2/libgit2@0b5ba0d744e69da5dc8c08d167c83dd87ed83af2 and libgit2/libgit2@c6184f0c4b209e462bf3f42ab20df2d13d8ee918.


lhchavez (approved):
oooh nice! i tried getting all of these at some point, and it seems like i didn't get all of them.

thanks!

PR #657, opened 2020-10-20
apply: Add bindings for git_apply_to_tree

SamiHiltunen:
Adds bindings to the git_apply_to_tree function that allows applying
a diff directly to a tree.

Should this get merged, when do you anticipate a new release that includes these changes?

### DIFF
--- a/diff.go
+++ b/diff.go
@@ -991,6 +991,24 @@ func (v *Repository) ApplyDiff(diff *Diff, location ApplyLocation, opts *ApplyOp
 	return nil
 }
 
+// ApplyToTree applies a Diff to a Tree and returns the resulting image as an Index.
+func (v *Repository) ApplyToTree(diff *Diff, tree *Tree, opts *ApplyOptions) (*Index, error) {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	var indexPtr *C.git_index
+	cOpts := opts.toC()
+	ecode := C.git_apply_to_tree(&indexPtr, v.ptr, tree.cast_ptr, diff.ptr, cOpts)
+	runtime.KeepAlive(diff)
+	runtime.KeepAlive(tree)
+	runtime.KeepAlive(cOpts)
+	if ecode != 0 {
+		return nil, MakeGitError(ecode)
+	}
+
+	return newIndexFromC(indexPtr, v), nil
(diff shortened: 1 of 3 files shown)


diff.go
```diff
@@ -991,6 +991,25 @@ func (v *Repository) ApplyDiff(diff *Diff, location ApplyLocation, opts *ApplyOp
...
 
+// ApplyToTree applies a Diff to a Tree and returns the resulting image as an Index.
+func (v *Repository) ApplyToTree(diff *Diff, tree *Tree, opts *ApplyOptions) (*Index, error) {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	var indexPtr *C.git_index
+	cOpts := opts.toC()
+	ecode := C.git_apply_to_tree(&indexPtr, v.ptr, tree.cast_ptr, diff.ptr, cOpts)
+	runtime.KeepAlive(v)
```
suhaibmujahid, line 1002: I do not think this is needed since you are using the variable at line `1010`.

SamiHiltunen: Good eye, removed. Thanks! 👍

diff_test.go
```diff
@@ -454,6 +455,135 @@ func TestApplyDiffAddfile(t *testing.T) {
...
 
+func TestApplyToTree(t *testing.T) {
+	repo := createTestRepo(t)
+	defer cleanupTestRepo(t, repo)
+
+	seedTestRepo(t, repo)
+
+	_, treeA := addAndGetTree(t, repo, "file", "a")
+	_, treeB := addAndGetTree(t, repo, "file", "b")
+	_, treeC := addAndGetTree(t, repo, "file", "c")
```
lhchavez, line 466: just a small request, can the commits and trees returned by this function be `defer ...Free()`d? I know that `TestApplyDiffAddFile()` doesn't do that either, but we should probably fix that later.

lhchavez (commented):
> Should this get merged, when do you anticipate a new release that includes these changes?

immediately! we have automation in place to cut a new path release the moment a new change gets merged.

diff_test.go
SamiHiltunen: @lhchavez done! 👍

lhchavez (approved):
:tada:

PR #658, opened 2020-10-22
repository: Implement wrappers for `git_object_lookup_prefix`

pks-t:
While we already have wrappers for `git_object_lookup`, there are none
yet for the prefixed variant where only the first n bytes of the OID are
used for the lookup. This commit adds them.


lhchavez (approved):
thanks!

PR #668, opened 2020-10-26
Create v31 to support libgit2 v1.1.x

suhaibmujahid:
The libgit2 `v1.1.0` is released two weeks ago. This PR allows `git2go` to link against the new version.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -7,11 +7,12 @@ on:
 jobs:
 
   backport:
+    name: Backport change to branch ${{ matrix.branch }}
+    continue-on-error: true
     strategy:
       fail-fast: false
       matrix:
-        branch: [ 'release-0.28', 'release-0.27' ]
-    name: Backport change to branch ${{ matrix.branch }}
+        branch: [ 'release-0.30', 'release-0.28', 'release-0.27' ]
 
     runs-on: ubuntu-20.04
 
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Due to the fact that Go 1.11 module versions have semantic meaning and don't nec
 | 0.28    | v28           |
 | 0.27    | v27           |
 
-You can import them in your project with the version's major number as a suffix. For example, if you have libgit2 v1.0 installed, you'd import git2go v31 with:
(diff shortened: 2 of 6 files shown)


lhchavez (commented):
Thanks! :D

suhaibmujahid:
Thank you for merging this.

I think the automatically generated version tag is not correct. 

See here: https://github.com/libgit2/git2go/releases/tag/v30.0.19

It should be v31 not v30. I think it should be deleted to not crash for ppl that depend on the older `libgit2` version.

Based on the documentation of the action that we are using:
https://github.com/anothrNick/github-tag-action#bumping
I supposed to include `#major` in the commit message to bump a major version.

lhchavez:
> Thank you for merging this.
> 
> I think the automatically generated version tag is not correct.
> 
> See here: https://github.com/libgit2/git2go/releases/tag/v30.0.19
> 
> It should be v31 not v30. I think it should be deleted to not crash for ppl that depend on the older `libgit2` version.
> 
> Based on the documentation of the action that we are using:
> https://github.com/anothrNick/github-tag-action#bumping
> I supposed to include `#major` in the commit message to bump a major version.

oops, thanks for noticing! new tags and the new `release-1.1` branch have been created.

suhaibmujahid:
Thank you for the fast response. 

I believe that the tag v30.0.19 should be deleted (or create v30.0.20 on the commit https://github.com/libgit2/git2go/commit/f83530b18dc46867ed06fc261b309b8b545a3b6f). Otherwise when someone updates or installs `v30`, by default will get v30.0.19 which depends on an unexpected `libgit2` version.

lhchavez:
> Thank you for the fast response.
> 
> I believe that the tag v30.0.19 should be deleted (or create v30.0.20 on the commit [f83530b](https://github.com/libgit2/git2go/commit/f83530b18dc46867ed06fc261b309b8b545a3b6f)). Otherwise when someone updates or installs `v30`, by default will get v30.0.19 which depends on an unexpected `libgit2` version.

the tag was also deleted when the v31 tag was created. the only thing left was the github release, which is not a real tag (but that's also removed now to avoid confusion)

PR #681, opened 2020-11-10
Add ReferenceNormalizeName

segevfiner:
Not sure about the buffer allocation for passing to `git_reference_normalize_name`: Is using `malloc` correct? Is the size correct, or should it be a loop with retries with some increasing buffer size (But the function doesn't seem to tell you how big it should be...)?

### DIFF
--- a/reference.go
+++ b/reference.go
@@ -489,6 +489,11 @@ func ReferenceIsValidName(name string) bool {
 	return false
 }
 
+const (
+	// This should match GIT_REFNAME_MAX in src/refs.h
+	refnameMaxLength = 1024
+)
+
 type ReferenceFormat uint
 
 const (
@@ -509,7 +514,7 @@ func ReferenceNormalizeName(name string, flags ReferenceFormat) (string, error) 
 	cname := C.CString(name)
 	defer C.free(unsafe.Pointer(cname))
 
-	bufSize := C.size_t(1024)
+	bufSize := C.size_t(refnameMaxLength)
 	buf := (*C.char)(C.malloc(bufSize))
 	defer C.free(unsafe.Pointer(buf))
 
--- a/reference.go
+++ b/reference.go
@@ -512,7 +512,7 @@ func ReferenceNormalizeName(name string, flags ReferenceFormat) (string, error) 
 	cname := C.CString(name)
 	defer C.free(unsafe.Pointer(cname))
 
-	bufSize := C.ulong(1024)
+	bufSize := C.size_t(1024)
 	buf := (*C.char)(C.malloc(bufSize))
 	defer C.free(unsafe.Pointer(buf))
 
--- a/reference.go
+++ b/reference.go
@@ -491,7 +491,7 @@ func ReferenceIsValidName(name string) bool {
 
 const (
 	// This should match GIT_REFNAME_MAX in src/refs.h
-	refnameMaxLength = 1024
+	_refnameMaxLength = C.size_t(1024)
 )
 
 type ReferenceFormat uint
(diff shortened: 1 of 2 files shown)


segevfiner:
Go 1.9?! Really... 🤷‍♂️

reference.go
```diff
@@ -488,3 +488,41 @@ func ReferenceIsValidName(name string) bool {
...
+
+// ReferenceNormalizeName normalizes the reference name and checks validity.
+//
+// This will normalize the reference name by removing any leading slash '/'
+// characters and collapsing runs of adjacent slashes between name components
+// into a single slash.
+//
+// Once normalized, if the reference name is valid, it will be returned in the
+// user allocated buffer.
+//
```
lhchavez, line 509: We can probably take this out since Go doesn't work that way.

```suggestion
```

reference.go
```diff
@@ -488,3 +488,41 @@ func ReferenceIsValidName(name string) bool {
...
+// See git_reference_symbolic_create() for rules about valid names.
+func ReferenceNormalizeName(name string, flags ReferenceFormat) (string, error) {
+	cname := C.CString(name)
+	defer C.free(unsafe.Pointer(cname))
+
+	bufSize := C.size_t(1024)
```
lhchavez, line 515: since `C.GIT_REFNAME_MAX` isn't exported, could that be explicitly declared?

```go
const (
    ...
    // This should match GIT_REFNAME_MAX in src/refs.h
    refnameMaxLength = 1024
)

...
func ReferenceNormalizeName(name string, flags ReferenceFormat) (string, error) {
   ...
   bufSize := C.size_t(refnameMaxLength)
```

reference_test.go
```diff
@@ -224,6 +224,29 @@ func TestReferenceIsValidName(t *testing.T) {
...
 }
 
+func TestReferenceNormalizeName(t *testing.T) {
+	t.Parallel()
+
+	ref, err := ReferenceNormalizeName("refs/heads//master", ReferenceFormatNormal)
+	checkFatal(t, err)
+
+	if ref != "refs/heads/master" {
+		t.Errorf("refs/heads//master should be normalized correctly")
```
lhchavez, line 234: Go prefers to print the [want and got](https://golang.org/pkg/testing/) when failing tests:

```suggestion
		t.Errorf("ReferenceNormalizeName(%q) = %q; want %q", "refs/heads//master", ref, want)
```

Same in L241.

lhchavez (commented):
nice! just a few small requests.

reference.go
```diff
@@ -488,3 +488,43 @@ func ReferenceIsValidName(name string) bool {
...
+// This will normalize the reference name by removing any leading slash '/'
+// characters and collapsing runs of adjacent slashes between name components
+// into a single slash.
+//
+// See git_reference_symbolic_create() for rules about valid names.
+func ReferenceNormalizeName(name string, flags ReferenceFormat) (string, error) {
+	cname := C.CString(name)
+	defer C.free(unsafe.Pointer(cname))
+
+	bufSize := C.size_t(refnameMaxLength)
```
suhaibmujahid, line 517: Since this is not gonna change, you could have it as a const.

```go
const (
	// This should match GIT_REFNAME_MAX in src/refs.h
	refnameMaxLength = 1024
	refnameBufSize = C.size_t(refnameMaxLength)
)
```

or just have it as:

```go
const (
	// This should match GIT_REFNAME_MAX in src/refs.h
	refnameBufSize = C.size_t(1024)
)
```

suhaibmujahid (commented):
Thank you for the PR. I have a trivial styling comment.

It is good to [prefix unexported globals with `_`](https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_):
```go
const (
	// This should match GIT_REFNAME_MAX in src/refs.h
	_refnameMaxLength = 1024
)
```

lhchavez (approved):

Issue #695, opened 2020-11-28
libgit2 version check


nmeum:
During compilation git2go performs the following libgit2 version check:

https://github.com/libgit2/git2go/blob/ad3ec3664d54779c4c2e49e41f85e886fbff343c/git_system_dynamic.go#L10-L12

Suppose I want to install a go software which uses git2go through `go get` on both Ubuntu 20.04 and Arch Linux I would run into the following problem: These Linux distributions ship different versions of libgit2. Since the version check performed by git2go is very strict (requiring an exact match of both minor and major version) the aforementioned go software would only compile with one specific version of libgit2, according to the version suffix I used for the git2go import.

IMHO it would be desirable if the git2go version check would be less strict. If I am informed correctly, libgit2 uses semantic versioning. As such, the API should not change in backwards incompatible ways without a major version bump. For instance, `github.com/libgit2/git2go/v30` (which currently requires libgit2 `1.0.0`) should IMHO compile with any libgit2 version where `LIBGIT2_VER_MAJOR == 1` and `LIBGIT2_VER_MINOR >= 0` (e.g. should compile fine with both libgit2 `1.1.0` and libgit2 `1.0.0`).

lhchavez:
sounds reasonable, although the important word here is _should_ (as in all v1.x.w _should_ work with any other v1.y.z where x < y).

I guess that if there's enough coverage for this in the CI, there's no reason not to give it a try. if we encounter a scenario where this is not true, we can replace the `LIBGIT2_VER_MINOR >= X` with `(LIBGIT2_VER_MINOR >= X && LIBGIT_VER_MINOR < Y)`, which still allows for some amount of leeway.

nmeum:
I think the required change is also very straightforward. I can just send a corresponding PR if you want me to do. Should the version check just be changed on the master branch for now or also on the release branches?

lhchavez:
if you can do the CI change (the most important although difficult and time-consuming part) together with the corresponding code change it would be preferable!

nmeum:
Unfortunately I am not too familiar with GitHub CI.

lhchavez:
they're pretty easy to work with! here's our full config: https://github.com/libgit2/git2go/blob/master/.github/workflows/ci.yml

essentially most of the work needed is to create / modify some shell scripts and then invoke the right Makefile command to run the tests.

nmeum:
And essentially you would want to modify the CI in a way that it uses different versions of libgit2 for testing?

[...]

PR #696, opened 2020-11-28
relax version check

nmeum:
Fixes: #695

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,8 @@ jobs:
   build-system-dynamic:
     strategy:
       fail-fast: false
+      matrix:
+        libgit2: [ '1.1.0' ]
     name: Go (system-wide, dynamic)
 
     runs-on: ubuntu-20.04
@@ -103,10 +105,10 @@ jobs:
       id: go
     - name: Check out code into the Go module directory
       uses: actions/checkout@v1
-    - name: Build libgit2
+    - name: Build libgit2 ${{ matrix.libgit2 }}
       run: |
         git submodule update --init
-        sudo ./script/build-libgit2.sh --dynamic --system
+        sudo env BUILD_LIBGIT_REF=v${{ matrix.libgit2 }} ./script/build-libgit2.sh --dynamic --system
     - name: Test
       run: make test
 
--- a/git_bundled_static.go
+++ b/git_bundled_static.go
@@ -9,8 +9,8 @@ package git
 #cgo CFLAGS: -DLIBGIT2_STATIC
 #include <git2.h>
 
-#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR != 1
-# error "Invalid libgit2 version; this git2go supports libgit2 v1.1"
+#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 1 || LIBGIT2_VER_MINOR > 1
+# error "Invalid libgit2 version; this git2go supports libgit2 between v1.1.0 and v1.1.0"
 #endif
 */
 import "C"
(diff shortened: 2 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -91,6 +91,8 @@ jobs:
   build-system-dynamic:
     strategy:
       fail-fast: false
+      matrix:
+        libgit2: [ '1.0.0', '1.0.1', '1.1.0' ]
```
lhchavez, line 95:
```suggestion
        libgit2: [ '1.1.0' ]
```

we probably only need to have a matrix with at most two versions: the current branch's version (in this case 1.1.0) and the largest known version number (in this case 1.1.0). patch versions only carry bugfixes with no API changes, so we can ignore those.

when this is cherry-picked to other branches (say, v30), it will need to be changed to

```yaml
        libgit2: [ '1.0.0', '1.1.0' ]
```

git_bundled_static.go
```diff
@@ -9,8 +9,8 @@ package git
 #cgo CFLAGS: -DLIBGIT2_STATIC
 #include <git2.h>
 
-#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR != 1
-# error "Invalid libgit2 version; this git2go supports libgit2 v1.1"
+#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 0
+# error "Invalid libgit2 version; this git2go supports libgit2 >= v1.0"
```
lhchavez, line 13: the test results show that this combination is broken, so we need to have this be:

```cpp
#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 1
# error "Invalid libgit2 version; this git2go supports libgit2 >= v1.1"
```

and since we need to update the CI anyways every time there's a new release to explicitly change the ref, might as well have this be:

```suggestion
#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 1 || LIBGIT2_VER_MINOR > 1
# error "Invalid libgit2 version; this git2go supports libgit2 between v1.1.0 and v1.1.0"
```

which is... exactly the same as what we have right now. BUT the difference is that in the other branches (say, v30) it'll be:

```cpp
#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 0 || LIBGIT2_VER_MINOR > 1
# error "Invalid libgit2 version; this git2go supports libgit2 between v1.0.0 and v1.1.0"
```

lhchavez (commented):
thanks for doing this, it looks good overall! we just need to make the CI green and we're good to merge :rocket:

git_bundled_static.go
nmeum: I don't think that an upper bound on the supported minor version is really strictly needed right now, but committed it anyhow as you suggested.

> the test results show that this combination is broken

Yep, only had that in for testing if the build would fail.

.github/workflows/ci.yml
nmeum: Changed the version matrix.

lhchavez (approved):

nmeum:
How does backporting this to other branches work? Even if the commit is cherry-picked the supported libgit2 version in the CI matrix needs to be adjusted for the other release branches.

lhchavez:
> How does backporting this to other branches work? Even if the commit is cherry-picked the supported libgit2 version in the CI matrix needs to be adjusted for the other release branches.

normally that's done through a GitHub action (https://github.com/libgit2/git2go/runs/1468017967?check_suite_focus=true), but it seems like the merge conflicts prevented that from happening :/ so it'll need to be done by hand. i can do that tomorrow, or you can give it a try by running `git cherry-pick -x 1fabe95fb7275df980ff6ab03fb85eac91c5849d` from the `release-1.0` branch and submit a PR for that.

Issue #617, opened 2020-06-20
could not determine kind of name for C.git_buf_dispose


xiaoxinda:
error is :
./../pkg/mod/github.com/libgit2/git2go/v29@v29.0.2/branch.go:211:8: could not determine kind of name for C.git_buf_dispose

centos:
my libgit2 is :0.99.0 
# pkg-config --modversion libgit2
# 0.99.0

[root@b2041d067d45 gitProxy]# go env
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/root/.cache/go-build"
GOENV="/root/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/root/go"
GOPRIVATE=""
GOPROXY="https://goproxy.cn,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/root/go/src/gitProxy/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build524585300=/tmp/go-build -gno-record-gcc-switches"

lhchavez:
hmm I'm unable to reproduce this :S 

```shell
cat > Dockerfile <<EOF
FROM centos:latest

RUN yum -y groupinstall "Development Tools"
RUN yum -y install cmake openssl-devel
RUN curl --location https://dl.google.com/go/go1.14.4.linux-amd64.tar.gz | \
    tar xz -C /usr/local
RUN curl --location https://github.com/libgit2/libgit2/releases/download/v0.99.0/libgit2-0.99.0.tar.gz | \
    tar xz -C ~
RUN mkdir -p ~/libgit2-0.99.0/build
RUN (cd ~/libgit2-0.99.0/build && \
     cmake \
[...]

PR #697, opened 2020-11-29
[release-1.0] Relax libgit2 minor version check (#696)

nmeum:
Backport of #696.

----

The major version must still be an exact match since libgit2 uses
semantic versioning and changes to the major number indicate backwards
incompatible changes to the API.

Fixes: #695
(cherry picked from commit 1fabe95fb7275df980ff6ab03fb85eac91c5849d)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,8 @@ jobs:
   build-system-dynamic:
     strategy:
       fail-fast: false
+      matrix:
+        libgit2: [ '1.0.0', '1.0.1', '1.1.0' ]
     name: Go (system-wide, dynamic)
 
     runs-on: ubuntu-20.04
@@ -103,10 +105,10 @@ jobs:
       id: go
     - name: Check out code into the Go module directory
       uses: actions/checkout@v1
-    - name: Build libgit2
+    - name: Build libgit2 ${{ matrix.libgit2 }}
       run: |
         git submodule update --init
-        sudo ./script/build-libgit2.sh --dynamic --system
+        sudo env BUILD_LIBGIT_REF=v${{ matrix.libgit2 }} ./script/build-libgit2.sh --dynamic --system
     - name: Test
       run: make test
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -92,7 +92,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        libgit2: [ '1.0.0', '1.0.1', '1.1.0' ]
+        libgit2: [ '1.0.0', '1.1.0' ]
     name: Go (system-wide, dynamic)
 
     runs-on: ubuntu-20.04
--- a/git_bundled_static.go
+++ b/git_bundled_static.go
@@ -9,8 +9,8 @@ package git
 #cgo CFLAGS: -DLIBGIT2_STATIC
 #include <git2.h>
 
-#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 0
-# error "Invalid libgit2 version; this git2go supports libgit2 >= v1.0"
+#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 0 || LIBGIT2_VER_MINOR > 1
+# error "Invalid libgit2 version; this git2go supports libgit2 between v1.0.0 and v1.1.0".
 #endif
 */
 import "C"
--- a/git_bundled_static.go
+++ b/git_bundled_static.go
@@ -9,8 +9,8 @@ package git
 #cgo CFLAGS: -DLIBGIT2_STATIC
 #include <git2.h>
 
-#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR != 0
-# error "Invalid libgit2 version; this git2go supports libgit2 v1.0"
+#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 0
+# error "Invalid libgit2 version; this git2go supports libgit2 >= v1.0"
 #endif
 */
 import "C"
--- a/git_system_dynamic.go
+++ b/git_system_dynamic.go
@@ -7,8 +7,8 @@ package git
 #cgo CFLAGS: -DLIBGIT2_DYNAMIC
 #include <git2.h>
 
-#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 0
-# error "Invalid libgit2 version; this git2go supports libgit2 >= v1.0"
+#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 0 || LIBGIT2_VER_MINOR > 1
+# error "Invalid libgit2 version; this git2go supports libgit2 between v1.0.0 and v1.1.0".
 #endif
 */
 import "C"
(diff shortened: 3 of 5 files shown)


git_bundled_static.go
```diff
@@ -9,8 +9,8 @@ package git
 #cgo CFLAGS: -DLIBGIT2_STATIC
 #include <git2.h>
 
-#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR != 0
-# error "Invalid libgit2 version; this git2go supports libgit2 v1.0"
+#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 0
+# error "Invalid libgit2 version; this git2go supports libgit2 >= v1.0"
```
lhchavez, line 13: Can these be written in the same style as the master change?

```suggestion
#if LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 0 || LIBGIT2_VER_MINOR > 1
# error "Invalid libgit2 version; this git2go supports libgit2 between v1.0.0 and v1.1.0".
```

.github/workflows/ci.yml
```diff
@@ -91,6 +91,8 @@ jobs:
   build-system-dynamic:
     strategy:
       fail-fast: false
+      matrix:
+        libgit2: [ '1.0.0', '1.0.1', '1.1.0' ]
```
lhchavez, line 95:
```suggestion
        libgit2: [ '1.0.0', '1.1.0' ]
```

we only need to consider the endpoints of the supported revisions to avoid overstraining the CI.

lhchavez (commented):
some differences from the master change need to be removed and we're ready to merge!

git_bundled_static.go
nmeum: I don't think that it is wise to have an upper bound on the supported minor version, having one really defeats the purpose of forward compatibility IMHO. Let's assume we include a upper bound check as you suggest: If libgit2 1.2.0 is released next week we need to adjust the version check again and all go software vendoring an old version of git2go will need to update git2go in order to build with libgit 1.2.0 even if libgit2 1.2.0 is compatible with the vendored git2go (which it should be due to semantic versioning).

.github/workflows/ci.yml
nmeum: Sure

git_bundled_static.go
lhchavez: it is similarly unwise to release untested combinations of software :(

semantic versioning is not perfect, unfortunately, and sometimes seemingly compatible changes are introduced that are not compatible with Go. we need to be defensive to avoid releasing any tag that is not backwards-compatible because that breaks the Go compatibility promises.

the good thing is that once new revisions of libgit2 are released, we just need to check that it's indeed compatible with the current branch and then tweak this version bound. we release a new tag and then everybody is happy since they can still refer to `"github.com/libgit2/git2go/v31"` (or whatever version they use) and be assured that the system library has been tested and works with their software.

nmeum: The problem I have with this is that I don't want to update the git2go version I am using all the time just for an increment in the version check. I really believe that this unnecessarily complicates installation of software which use git2go as I can't tell my users to simply install libgit2 1.X as a dependency. Every time a new libgit2 is released and I update my distro packages my build basically breaks. If you really insist on a strict minor version check would it be possible to introduce some kind of option for disabling it?

lhchavez: new versions are released very infrequently (1-2 per year), so the overhead would not be as large. we already need to do some work around those times (like create new branches, bump versions, etc.) so adding this extra task of tweaking the minor version ranges for the previous release branches is palatable for us.

that way we won't have the build breaking for the opposite problem: a version of libgit2 1.x that should have been compatible that for some reason isn't.

regarding adding an option for disabling the version check, how would it look like? AFAIK the only way to achieve this would be to add a new tag, but that would defeat the purpose of being `go get` compatible.

nmeum: > […] tweaking the minor version ranges for the previous release branches is palatable for us.

Sure, but it still annoying for downstream consumers to update git2go all the time ;)

> that way we won't have the build breaking for the opposite problem: a version of libgit2 1.x that should have been compatible that for some reason isn't.

I guess I have more faith in the semantic versioning scheme here, breakage introduced in a minor revision should IMHO be considered a bug in libgit2 and not in git2go. That's really the beauty of correctly executed semantic versioning, but as you pointed out there is no guarantee that a minor revision might not introduce a regression, but then again there is also no guarantee that a minor version didn't introduce a regression just because the git2go test suite passes.

Furthermore, even if there is a regression it might not even be reachable in the package which imports git2go (e.g. because it doesn't use the function which causes the regression). Why not leave version checking to developers importing git2go just like libgit2 itself does?

> regarding adding an option for disabling the version check, how would it look like?

Best I came up with so far would be something like:

```C
#if (LIBGIT2_VER_MAJOR != 1 || LIBGIT2_VER_MINOR < 0 || LIBGIT2_VER_MINOR > 1) && !defined(GIT2GO_SKIP_VER_CHECK)
# error "Invalid libgit2 version; this git2go supports libgit2 between v1.0.0 and v1.1.0".
#endif
```

and then invoke `go get` as:

```
CGO_CFLAGS=-DGIT2GO_SKIP_VER_CHECK  go get <…>
```

Would be even better if a package could set `CFLAGS` for other packages it is importing, thereby eliminating the need to manually set the environment variable, but I don't think that doing so is possible.

We can also turn the `#error` into a `#warning` and add a `!defined(…)` to disable the warning :thinking:

lhchavez: Something else to consider is that Go's versioning philosophy is not _entirely_ in-line with semantic versioning, aside from the fact that breaking incompatibilities are marked with a major version bump, but still recognizes that people will want to use semantic versioning, so it arrived at a sort of trade-off for supporting parts of semver. One of Go's language creators wrote an amazing series of articles about Go's versioning philosophy and the rationales behind it (https://research.swtch.com/vgo-principles is the last one and is a sort of TL;DR of the whole thing), and in one argues that in order to maintain [build repeatability](https://research.swtch.com/vgo-repro), when multiple libraries depend on different version of another library, the [minimum version](https://research.swtch.com/vgo-mvs) that satisfies all constraints should be chosen, instead of the more usual choice of picking the largest possible that satisfies semantic versioning or using SAT solvers. There's also the fact that even if the API doesn't change between minor versions, semantics might inadvertently change even when just fixing a bug (see [Hyrum's Law](http://hyrumslaw.com/)).

One consequence of this is that library authors are expected to bump their dependency versions (and thus test their software with that particular version combination) more often, so that the minimum version advances as much as possible. libgit2 is not subject to such strict rules, so breaking Go's compatibility expectations is definitely not a bug in libgit2. Therefore, this wrapper library needs to be a bit more careful about it and provide an abstraction that respects the conventions of Go and therefore the expectations of users. That requirement to bump versions more often is definitely a burden, but overall the way versioning works does have its benefits.

I think the solution achieved in this change is a pretty good trade-off. It still goes with Go's versioning philosophy's spirit of explicitly checking for compatibility (although as you point out, it's not perfect, nor can we expect it to ever be) while providing some amount of flexibility and ease of use that was previously not possible. Even with the trade-offs, it's definitely a net-positive for everyone, so thanks for doing it!

w.r.t. the option to disable the version check, since it's explicitly opt-in, it might be okay to consider in the future as an escape hatch if the current solution ends up not panning out for one reason or another (no `#warning` needed: users are expected to be familiar with how the build works by opting into using non-standard flags in case an [inscrutable build error](https://github.com/libgit2/git2go/issues/617) appears).

Once again, thanks for this change! :D

lhchavez (approved):
:rocket:

PR #728, opened 2020-12-15
Add support for custom smart transports

lhchavez:
This change adds support for git smart transports. This will be then
used to implement http, https, and ssh transports that don't rely on the
libgit2 library.


suhaibmujahid:
This looks cool. Thank you.

lhchavez:
Superseded by https://github.com/libgit2/git2go/pull/806

Issue #467, opened 2018-11-15
Reflog support


bk2204:
libgit2 provides support for reading and iterating over reflogs.  However, git2go doesn't seem to provide this support.  Is it possible this support could be added?

My particular use case, if it matters, is wanting to know when the last fetch in a git2go-controlled repository occurred.

PR #730, opened 2021-01-18
Reflog support (#467)

pokstad:
Adds support and tests for all reflog functions listed on https://libgit2.org/libgit2/#HEAD/group/reflog.

### DIFF
--- a/reflog.go
+++ b/reflog.go
@@ -1,0 +1,175 @@
+package git
+
+/*
+#include <git2.h>
+*/
+import "C"
+import (
+	"runtime"
+	"unsafe"
+)
+
+// Reflog is a log of changes for a reference
+type Reflog struct {
+	ptr  *C.git_reflog
+	repo *Repository
+	name string
+}
+
+func newRefLogFromC(ptr *C.git_reflog, repo *Repository, name string) *Reflog {
+	l := &Reflog{
+		ptr:  ptr,
+		repo: repo,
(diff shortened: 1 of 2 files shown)


reflog.go
```diff
@@ -0,0 +1,175 @@
...
+	runtime.KeepAlive(l)
+
+	return goEntry
+}
+
+func (l *Reflog) DropEntry(index uint, rewriteHistory bool) error {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	var rewriteHistoryInt int
```
lhchavez, line 139:
```suggestion
	var rewriteHistoryInt C.int
```

might as well define it as `C.int` from the get-go to avoid the cast in L144.

reflog.go
```diff
@@ -0,0 +1,175 @@
...
+	cSignature, err := committer.toC()
+	if err != nil {
+		return err
+	}
+	defer C.git_signature_free(cSignature)
+
+	cMsg := C.CString(message)
+	defer C.free(unsafe.Pointer(cMsg))
+
+	C.git_reflog_append(l.ptr, oid.toC(), cSignature, cMsg)
```
lhchavez, line 166: can the error be handled?

lhchavez (commented):
sorry for the delay! for some reason i saw this, accidentally marked as read and completely forgot about it until today u_u

PR #733, opened 2021-02-01
Support git_remote_create_with_opts

### LINKED ISSUE #645: Support git_remote_create_with_opts
Hello, first of all, thanks for such a great project!

It would be nice to support  https://libgit2.org/libgit2/#HEAD/group/remote/git_remote_create_with_opts , which is introduced in v0.28.0 . Thank you!

bc-lee:
Closes #645

### DIFF
--- a/remote.go
+++ b/remote.go
@@ -556,6 +556,7 @@ func (c *RemoteCollection) Create(name string, url string) (*Remote, error) {
 	return remote, nil
 }
 
+//CreateWithOptions Creates a repository object with extended options.
 func (c *RemoteCollection) CreateWithOptions(url string, option *RemoteCreateOptions) (*Remote, error) {
 	remote := &Remote{repo: c.repo}
 
--- a/remote.go
+++ b/remote.go
@@ -18,21 +18,21 @@ import (
 	"unsafe"
 )
 
-type RemoteCreate uint
+// RemoteCreateOptionsFlag is Remote creation options flags
+type RemoteCreateOptionsFlag uint
 
 const (
 	// Ignore the repository apply.insteadOf configuration
-	RemoteCreateSkipInsteadof RemoteCreate = C.GIT_REMOTE_CREATE_SKIP_INSTEADOF
+	RemoteCreateSkipInsteadof RemoteCreateOptionsFlag = C.GIT_REMOTE_CREATE_SKIP_INSTEADOF
 	// Don't build a fetchspec from the name if none is set
-	RemoteCreateSkipDefaultFetchspec RemoteCreate = C.GIT_REMOTE_CREATE_SKIP_DEFAULT_FETCHSPEC
+	RemoteCreateSkipDefaultFetchspec RemoteCreateOptionsFlag = C.GIT_REMOTE_CREATE_SKIP_DEFAULT_FETCHSPEC
 )
 
 // RemoteCreateOptions contains options for creating a remote
 type RemoteCreateOptions struct {
-	Repository *Repository
 	Name       string
 	FetchSpec  string
-	Flags      RemoteCreate
+	Flags      RemoteCreateOptionsFlag
 }
 
 type TransferProgress struct {
@@ -556,7 +556,7 @@ func (c *RemoteCollection) Create(name string, url string) (*Remote, error) {
 	return remote, nil
 }
 
-func (c *RemoteCollection) CreateOptions(url string, option *RemoteCreateOptions) (*Remote, error) {
+func (c *RemoteCollection) CreateWithOptions(url string, option *RemoteCreateOptions) (*Remote, error) {
 	remote := &Remote{repo: c.repo}
 
 	curl := C.CString(url)
@@ -565,9 +565,10 @@ func (c *RemoteCollection) CreateOptions(url string, option *RemoteCreateOptions
 	runtime.LockOSThread()
 	defer runtime.UnlockOSThread()
 
-	opts :=	remoteCreateOptionsToC(option)
+	opts :=	populateRemoteCreateOptions(c.repo, option)
 	defer freeRemoteCreateOptions(opts)
 	ret := C.git_remote_create_with_opts(&remote.ptr, curl, opts)
+	runtime.KeepAlive(c.repo)
 	if ret < 0 {
 		return nil, MakeGitError(ret)
 	}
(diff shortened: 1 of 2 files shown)


remote.go
```diff
@@ -18,6 +18,23 @@ import (
 	"unsafe"
 )
 
+type RemoteCreate uint
```
lhchavez, line 21: could this be

```suggestion
type RemoteCreateOptionsFlag uint
```

for consistency with https://godoc.org/github.com/libgit2/git2go#BlameOptionsFlag ?  bonus points if a docstring is added.

remote.go
```diff
@@ -18,6 +18,23 @@ import (
...
+const (
+	// Ignore the repository apply.insteadOf configuration
+	RemoteCreateSkipInsteadof RemoteCreate = C.GIT_REMOTE_CREATE_SKIP_INSTEADOF
+	// Don't build a fetchspec from the name if none is set
+	RemoteCreateSkipDefaultFetchspec RemoteCreate = C.GIT_REMOTE_CREATE_SKIP_DEFAULT_FETCHSPEC
+)
+
+// RemoteCreateOptions contains options for creating a remote
+type RemoteCreateOptions struct {
+	Repository *Repository
```
lhchavez, line 32:
```suggestion
```

it's probably better to not have this be passed in through the `Options` object, since the pointer to the C repository can be obtained from within `CreateWithOptions`, which has a strong affinity to a pre-existing `git.Repository` and would be confusing (API-wise) that one would be able to call `myOneRepository.Remotes.CreateWithOptions(...)` and the resulting remote would be associated with a repository that is not `myOneRepository`.

in order to support detached remotes, we could introduce `func NewDetachedRemote(url string, options *RemoteCreateOptions) (*Remote, error)`.

remote.go
```diff
@@ -539,6 +556,26 @@ func (c *RemoteCollection) Create(name string, url string) (*Remote, error) {
 	return remote, nil
 }
 
+func (c *RemoteCollection) CreateOptions(url string, option *RemoteCreateOptions) (*Remote, error) {
```
lhchavez, line 559:
```suggestion
func (c *RemoteCollection) CreateWithOptions(url string, option *RemoteCreateOptions) (*Remote, error) {
```

for consistency with the `_with_` part of the C function.

also, can there be a docstring for this function?

remote.go
```diff
@@ -1027,3 +1064,46 @@ func (o *Remote) Prune(callbacks *RemoteCallbacks) error {
...
+	opts := C.git_remote_create_options{}
+	ecode := C.git_remote_create_options_init(&opts, C.GIT_REMOTE_CREATE_OPTIONS_VERSION)
+	if int(ecode) != 0 {
+		return nil, MakeGitError(ecode)
+	}
+
+	return &RemoteCreateOptions{
+		Repository: nil,
+		Flags:      0,
+	}, nil
```
lhchavez, line 1082:
```suggestion
	return &RemoteCreateOptions{
		Flags:      RemoteCreateOptionsFlag(opts.flags),
	}, nil
```

otherwise `opts` is unused!

remote.go
```diff
@@ -1027,3 +1064,46 @@ func (o *Remote) Prune(callbacks *RemoteCallbacks) error {
...
+}
+
+func remoteCreateOptionsToC(opts *RemoteCreateOptions) (copts *C.git_remote_create_options) {
```
lhchavez, line 1085: could this follow the `populateXxxOptions` convention found in other files? e.g. https://github.com/libgit2/git2go/blob/4b2ac7c998be677d865367908787f17fb570c679/revert.go#L18

remote.go
```diff
@@ -539,6 +556,26 @@ func (c *RemoteCollection) Create(name string, url string) (*Remote, error) {
...
+
+	curl := C.CString(url)
+	defer C.free(unsafe.Pointer(curl))
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	opts :=	remoteCreateOptionsToC(option)
+	defer freeRemoteCreateOptions(opts)
+	ret := C.git_remote_create_with_opts(&remote.ptr, curl, opts)
```
lhchavez, line 570:
```suggestion
	ret := C.git_remote_create_with_opts(&remote.ptr, curl, opts)
	runtime.KeepAlive(c.repo)
```

otherwise, the go runtime is completely free to free the repository under us while the Cgo call is being made.

lhchavez (commented):
nice, thanks for doing this!

remote.go
bc-lee: In my opinion, `NewDetachedRemote` is not needed right now. That was my design mistake. (Due to a dead copy of the C interface).

remote.go
```diff
@@ -1027,3 +1066,45 @@ func (o *Remote) Prune(callbacks *RemoteCallbacks) error {
...
+	if int(ecode) != 0 {
+		return nil, MakeGitError(ecode)
+	}
+
+	return &RemoteCreateOptions{
+		Flags: RemoteCreateOptionsFlag(opts.flags),
+	}, nil
+}
+
+func populateRemoteCreateOptions(repository *Repository, opts *RemoteCreateOptions) (copts *C.git_remote_create_options) {
```
lhchavez, line 1086:
```suggestion
func populateRemoteCreateOptions(copts *C.git_remote_create_options, opts *RemoteCreateOptionsrepository *Repository) *C.git_remote_create_options {
```

for consistency with all the other `populateXxxOptions()` functions.

remote.go
```diff
@@ -1027,3 +1066,45 @@ func (o *Remote) Prune(callbacks *RemoteCallbacks) error {
...
+	if repository == nil || opts == nil {
```
lhchavez, line 1087:
```suggestion
	if opts == nil {
```

having a nil `repository` is completely legal.

remote.go
```diff
@@ -1027,3 +1066,45 @@ func (o *Remote) Prune(callbacks *RemoteCallbacks) error {
...
 }
+
+// DefaultApplyOptions returns default options for remote create
+func DefaultRemoteCreateOptions() (*RemoteCreateOptions, error) {
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	opts := C.git_remote_create_options{}
+	ecode := C.git_remote_create_options_init(&opts, C.GIT_REMOTE_CREATE_OPTIONS_VERSION)
+	if int(ecode) != 0 {
```
lhchavez, line 1077:
```suggestion
	if ecode < 0 {
```

for consistency with all other places `ecode` is used.

lhchavez (commented):
just a few more style requests and we're good!

lhchavez (approved):
thanks :D

PR #735, opened 2021-02-01
Implement git_repository_set_config

### LINKED ISSUE #732: Support git_repository_set_config
It would be nice to support https://github.com/libgit2/libgit2/blob/v1.1.0/include/git2/sys/repository.h#L86 which is introduced in v0.16.0 . Thank you!

bc-lee:
Closes #732

### DIFF
--- a/repository.go
+++ b/repository.go
@@ -165,6 +165,15 @@ func (v *Repository) Config() (*Config, error) {
 	return config, nil
 }
 
+func (v *Repository) SetConfig(c *Config) error {
+	ret := C.git_repository_set_config(v.ptr, c.ptr)
+	runtime.KeepAlive(v)
+	if ret < 0 {
+		return MakeGitError(ret)
+	}
+	return nil
+}
+
 func (v *Repository) Index() (*Index, error) {
 	var ptr *C.git_index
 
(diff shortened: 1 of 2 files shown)


repository.go
```diff
@@ -165,6 +165,18 @@ func (v *Repository) Config() (*Config, error) {
 	return config, nil
 }
 
+func (v *Repository) SetConfig(c *Config) error {
```
lhchavez, line 168:
```suggestion
// SetConfig sets the configuration file for this repository.
//
// This configuration file will be used for all configuration queries involving
// this repository.
func (v *Repository) SetConfig(c *Config) error {
```

repository.go
```diff
@@ -165,6 +165,18 @@ func (v *Repository) Config() (*Config, error) {
...
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ret := C.git_repository_set_config(v.ptr, c.ptr)
+	runtime.KeepAlive(v)
```
lhchavez, line 173:
```suggestion
	runtime.KeepAlive(v)
	runtime.KeepAlive(c)
```

otherwise if `c` is not used by the caller later, the Go runtime is free to remove it in the middle of the CGO call.

lhchavez (approved):

PR #750, opened 2021-02-15
Make index time fields public

### LINKED ISSUE #304: IndexTime fields are private, how to I create IndexTime{}?
I'm trying to add an Entry to an Index:

``` golang
        // Create git blob from file
        blobOid, err := repo.CreateBlobFromBuffer(file.Content)
        if err != nil {
            return err
        }

        // Create git index entry
        indexEntry := &git.IndexEntry{
            // Since v23 IndexTime{} is required. Which is not possible to create from here, as all
            // fields are private. See https://github.com/libgit2/git2go/blob/master/index.go
            // 
            // Ctime: git.IndexTime{}.
            // Mtime: git.IndexTime{},
            Mode:  git.FilemodeBlob,
            Uid:   uint32(os.Getuid()),
            Gid:   uint32(os.Getgid()),
            Size:  uint32(len(file.Content)),
            Id:    blobOid,
            Path:  file.Name,
        }

        // Add index to git tree
        if err := index.Add(indexEntry); err != nil {
            return err
        }
```

Unfortunately, since git2go v23, I cannot set `Ctime` and `Mtime`, ad they are requiring an IndexTime element, which I cannot fill with e.g. the current time, as the fields of `IndexTime` are private. 
Is this an bug, or am I supposed to create the `IndexEntry` in another way? I couldn't find a `NewIndexEntry()` function or any of the sort.

I'd apprechiate a hint!

mbfr:
From gorelease:
```
Compatible changes:
- IndexTime.Nanoseconds: added
- IndexTime.Seconds: added
```
There are no extra tests because there isn't really anything to test

closes #304

### DIFF
--- a/index.go
+++ b/index.go
@@ -57,8 +57,8 @@ type Index struct {
 }
 
 type IndexTime struct {
-	seconds     int32
-	nanoseconds uint32
+	Seconds     int32
+	Nanoseconds uint32
 }
 
 type IndexEntry struct {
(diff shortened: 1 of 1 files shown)


lhchavez (approved):
thanks! :D

PR #755, opened 2021-03-02
initial commit on adding usage of git_create_remote_detached

ralucas:
This creates a `RemoteDetached` type that is instantiable with a remote uri string.  It has only one method, `LsDetached` for emulating `git ls-remote <remote-uri>`.

The reason for this is that it looks like there is no other way to simply do a `git ls-remote` type request without creating a repository reference locally.

Thank you for your time in reviewing this.


remote_test.go
```diff
@@ -493,3 +495,40 @@ func TestRemoteSSH(t *testing.T) {
...
+
+func TestNewRemoteDetached(t *testing.T) {
+	_, err := NewRemoteDetached(TestGitRepoUri)
+	if err != nil {
+		t.Fatalf("error %v", err)
+	}
+}
+
+func TestRemoteDetached_LsDetached(t *testing.T) {
+	r, _ := NewRemoteDetached(TestGitRepoUri)
```
lhchavez, line 507:
```suggestion
	remote, err := NewRemoteDetached(TestGitRepoUri)
	checkFatal(t, err)
	defer remote.Free()
```

for consistency with other functions that create remotes (and to avoid memory leaks)

remote.go
```diff
@@ -173,6 +173,37 @@ type Remote struct {
...
+
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+
+	ret := C.git_remote_create_detached(&ptr, C.CString(remoteUri))
+	if ret != 0 {
+		return nil, MakeGitError(ret)
+	}
+
+	return &RemoteDetached{r: Remote{ptr: ptr}}, nil
```
lhchavez, line 193: would it be possible for this to return the `*Remote` directly? That way the `RemoteDetached` and the `LsDetached` shouldn't be needed, since callers can interact with the remote using the normal API, which will avoid having to implement more APIs on `RemoteDetached` that would only forward arguments to the underlying `Remote`. It's totally fine if the `repo` property is `nil`: it's just there to avoid the GC from deleting the `Repository` when the `Remote` is still around, but here it will be `nil`.

also, the finalizer for the `Remote` object must be set to avoid memory leaks:

https://github.com/libgit2/git2go/blob/75708eea9844a3ae7cd48ace71d50fe8fbfc91c1/remote.go#L586

lhchavez (commented):
thanks for the change!

PR #757, opened 2021-04-03
Git repository item path

vladimir-buzuev:
add wrapper for `git_repository_item_path`


lhchavez (commented):
looks good! would it be possible to run `go fmt` on repository.go? that way the next commit that runs that command won't overwrite L748-L671 unnecessarily.

lhchavez (approved):
thanks!

Issue #543, opened 2020-02-29
Stringer for ErrClass and ErrCode ?


slyphon:
The functionality might already exist somewhere, but it seems like it'd be useful for error reporting to make a stringer for ErrClass and ErrCode to return human-readable names for the corresponding constants.  

It'd mean that `go generate` would have to be run when the relevant constants were changed, but it seems like this should be relatively easy to do.

Would you accept a pull request that set this up?

lhchavez:
there's precedent for that (e.g. #523) so PR away!

now that you mention it, it may be worth adding an Action to check whether all the `go generate`d files are up to date. that we can do separately.

PR #841, opened 2021-10-14
Generate stringer files automatically

g4s8:
Added `stringer` annotations to `git.go` for `ErrorClass` and
`ErrorCode`. Added `generate` rule for `Makefile` to generate
string representations for these types (first building cgo files
in `_obj` dir to get C constants). Finally, updated `ci` actions
workflow to check that generated files are up to date.

Fixes: #543

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,4 +120,4 @@ jobs:
     - name: Generate files
       run: make generate
     - name: Check nothing changed
-      run: git diff --quiet --exit-code || (echo "detected changes after generate" && exit 1)
+      run: git diff --quiet --exit-code || (echo "detected changes after generate" ; git status ; exit 1)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,6 +121,7 @@ jobs:
       run: |
         git submodule update --init
         sudo apt-get install -y --no-install-recommends libssh2-1-dev
+        go install golang.org/x/tools/cmd/stringer@latest
     - name: Generate files
       run: make generate
     - name: Check nothing changed
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,6 +117,10 @@ jobs:
       id: go
     - name: Check out code into the Go module directory
       uses: actions/checkout@v2
+    - name: Install libgit2 build dependencies
+      run: |
+        git submodule update --init
+        sudo apt-get install -y --no-install-recommends libssh2-1-dev
     - name: Generate files
       run: make generate
     - name: Check nothing changed
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,6 +123,8 @@ jobs:
         sudo apt-get install -y --no-install-recommends libssh2-1-dev
         go install golang.org/x/tools/cmd/stringer@latest
     - name: Generate files
-      run: make generate
+      run: |
+        export PATH=$(go env GOPATH)/bin:$PATH
+        make generate
     - name: Check nothing changed
       run: git diff --quiet --exit-code || (echo "detected changes after generate" ; git status ; exit 1)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,3 +105,19 @@ jobs:
         sudo ./script/build-libgit2.sh --static --system
     - name: Test
       run: go test --count=1 --tags "static,system_libgit2" ./...
+
+  check-generate:
+    name: Check generated files were not modified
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -105,3 +105,19 @@ jobs:
...
+      uses: actions/setup-go@v1
+      with:
+        go-version: '1.17'
+      id: go
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+    - name: Generate files
+      run: make generate
+    - name: Check nothing changed
+      run: git diff --quiet --exit-code || (echo "detected changes after generate" && exit 1)
```
lhchavez, line 123:
```suggestion
      run: git diff --quiet --exit-code || (echo "detected changes after generate" ; git status ; exit 1)
```

this way the caller will be able to see what were the changes.

Makefile
```diff
@@ -53,3 +53,17 @@ test-static: static-build/install/lib/libgit2.a
...
+	find ./_obj -type f ! -name '*.go' -exec rm {} \;
+	mv ./_obj/_cgo_gotypes.go ./_obj/cgo_gotypes.go
+	cd ./_obj && go generate ./...
+	find ./_obj -type f -name '*_string.go' -exec mv -v {} . \;
+	rm -fr ./_obj
+endef
+
+generate:
+	$(call _gen_file,diff.go)
+	$(call _gen_file,git.go)
```
lhchavez, line 69: is it possible to have this be

```suggestion
generate: static-build/install/lib/libgit2.a
	go generate --tags "static" ./...
```

? that way whenever a new stringer is added, the Makefile doesn't need to be updated (folks will forget about this, haha). the `_gen_file` function also depends a bit on implementation details that are not guaranteed to be true forever.

lhchavez (commented):
yay, thanks for the CI addition!

Makefile
g4s8: Unfortunately, `go generate --tags "static" ./...` will not work with CGO constants from libgit2 headers (as in `diff.go` and `git.go`), so it's a workaround to extract these constants from header files to `.go` files, see this question: https://stackoverflow.com/q/69566767/1723695

g4s8:
@lhchavez thanks for review! Please check new changes and answers.

Makefile
lhchavez: oh that's likely because the static library wasn't correctly built / installed (that's why the `generate:` rule has `static-build/install/lib/libgit2.a` as a dependency, which fixes this issue). i just tried this on my machine and it worked. can you give this a try?

also, minor thing, can the `generate` rule be moved to just below the `default:` one? i just realized that it's included in the various library sections, and that was a bit confusing, haha.

.github/workflows/ci.yml
```diff
@@ -105,3 +105,19 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Set up Go
+      uses: actions/setup-go@v1
+      with:
+        go-version: '1.17'
+      id: go
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+    - name: Generate files
```
lhchavez, line 120:
```suggestion
    - name: Install libgit2 build dependencies
      run: |
        git submodule update --init
        sudo apt-get install -y --no-install-recommends libssh2-1-dev
    - name: Generate files
```

oops, forgot this last round! otherwise `make generate` will fail because CGo won't be able to find `git2.h`.

Makefile
g4s8: Thanks, it's working now, updated in the latest commit

.github/workflows/ci.yml
```diff
@@ -105,3 +105,23 @@ jobs:
...
+      uses: actions/setup-go@v1
+      with:
+        go-version: '1.17'
+      id: go
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+    - name: Install libgit2 build dependencies
+      run: |
+        git submodule update --init
+        sudo apt-get install -y --no-install-recommends libssh2-1-dev
```
lhchavez, line 123:
```suggestion
        sudo apt-get install -y --no-install-recommends libssh2-1-dev
        go install golang.org/x/tools/cmd/stringer@latest
```

haha, `stringer` wasn't installed! (and we may need to tweak `PATH` unless GH does it for us)

g4s8:
Exported `GOPATH` in https://github.com/libgit2/git2go/pull/841/commits/9c737a6d50a84f16b4387ddd109ef0fbf4a49b5c

lhchavez (approved):
yay! now it works in CI!

thank you for this improvement :D

PR #872, opened 2021-12-22
Add ability to specify ProxyOptions for push operations

au2001:
Analog to https://github.com/libgit2/git2go/pull/623 but for push operations rather than fetch.


lhchavez (approved):
sorry for the delay and thanks for this improvement!

PR #886, opened 2022-01-23
readme: link to godoc for current main branch

dylrich:
The GoDoc link currently refers to the initial version of this package, but usually folks are looking
for whatever matches the current main branch. This should be updated whenever the package
version changes.


lhchavez (approved):
oh nice! thank you :D

PR #895, opened 2022-01-23
readme: link to godoc for current main branch

dylrich:
This is a release-specific change corresponding to #886


dylrich:
I noticed that this release didn't get an automatic pull request from the bot, so I'm not sure if this one needs an update. Thought I'd put it out there just in case.

PR #898, opened 2022-01-28
Add refspec bindings

wabain:
This adds support for the parse, accessor, and transform functions for [refspec](https://libgit2.org/libgit2/#v1.3.0/group/refspec) objects. It's a direct port of the C API.


lhchavez (approved):
Thanks for adding tests :D

